PR #46, opened 2018-10-17
Feature/oo ap1rog

lelemmen:
Bring over the Jacobi orbital optimization algorithm from the other library.


src/AP1roG/AP1roG.cpp
```diff
@@ -24,4 +24,35 @@ AP1roG::AP1roG(const GQCG::AP1roGGeminalCoefficients& geminal_coefficients, doub
...
+double calculateAP1roGEnergy(const GQCG::AP1roGGeminalCoefficients& G, const GQCG::HamiltonianParameters& ham_par) {
+
+    Eigen::MatrixXd h_SO = ham_par.get_h().get_matrix_representation();
+    Eigen::Tensor<double, 4> g_SO = ham_par.get_g().get_matrix_representation();
+
+
+    // KISS implementation of the AP1roG energy
+    double E = 0.0;
+    for (size_t j = 0; j < G.get_N_P(); j++) {
+        E += 2 * h_SO(j,j);
```
tmhuysen: Does h_SO(j,j) require to be h_SO(j, j) ? if so, repeat through entire code.

src/AP1roG/AP1roGGeminalCoefficients.cpp
```diff
@@ -68,4 +63,91 @@ Eigen::MatrixXd AP1roGGeminalCoefficients::asMatrix() const {
...
+ *
+ *      Note that
+ *          - i is in [0 ... N_P[       is the 'major' index (i.e. changes in i are not contiguous)
+ *          - a is in [N_P ... K[       is the 'minor' index (i.e. changes in a are contiguous)
+ */
+size_t AP1roGGeminalCoefficients::vectorIndex(size_t i, size_t a) const {
+
+    // Check for invalid values for i and a
+    if (i >= this->N_P) {
+        throw std::invalid_argument("i must be smaller than N_P.");
```
tmhuysen: add more context than "i"  e.g. ->  "Major index i"

src/AP1roG/AP1roGGeminalCoefficients.cpp
```diff
@@ -68,4 +63,91 @@ Eigen::MatrixXd AP1roGGeminalCoefficients::asMatrix() const {
...
+    }
+    if (a < this->N_P) {
+        throw std::invalid_argument("a must be larger than or equal to N_P.");
```
tmhuysen: idem

src/AP1roG/AP1roG.cpp
lelemmen: I don't think we have to add whitespace to element-accessors for matrices like here.

PR #61, opened 2018-10-18
Bug: elicit errant behavior from Travis.

guacke:
I have noticed that Travis errors when using different optimization flags. In this branch, we will try to fix this bug.


lelemmen:
Is this bug related to the `-O2 -march=native` flag?

guacke:
Yes. This could point to a 'hidden' bug in the code. According to the Travis logs, this should be related to a 'double free or corruption'.

lelemmen:
'double free' seems to mean that there is a double `free(x)` operation, so it could mean that we are trying to delete a `nullptr`.

- Looking at the [Travis log](https://travis-ci.org/GQCG/gqcg/builds/443118671?utm_source=github_status&utm_medium=notification), the issue doesn't seem to be with the `std::shared_ptr<GQCG::AOBasis`, as for example the RHF- and HamiltonianConstructor-tests do not fail.

The issue could be related to numopt, as all failing tests use numopt. I'm not sure the bad guy is numopt, because the CISolver-test doesn't fail.

lelemmen:
@tmhuysen Could you profile the code and check where the double free is coming from?

lelemmen:
Log from Clang 6.0.1:

Running 3 test cases...
Will read basis set from /usr/local/libint/2.3.1/share/libint/2.3.1/basis/sto-3g.g94
Will read basis set from /usr/local/libint/2.3.1/share/libint/2.3.1/basis/sto-3g.g94
Will read basis set from /usr/local/libint/2.3.1/share/libint/2.3.1/basis/6-31g**.g94
AP1roGPSESolver_test(2761,0x11ad1c5c0) malloc: *** error for object 0x7fd963d02620: pointer being freed was not allocated
AP1roGPSESolver_test(2761,0x11ad1c5c0) malloc: *** set a breakpoint in malloc_error_break to debug
unknown location:0: fatal error: in "h2_631gdp": signal: SIGABRT (application abort requested)
/Users/laurentlemmens/Software/gqcg/tests/AP1roG/AP1roGPSESolver_test.cpp:41: last checkpoint: "h2_631gdp" test entry

*** 1 failure is detected in the test module "AP1roGPSESolver"

tmhuysen:
This bug has been solved, and was added as a separate feature.

PR #105, opened 2018-11-01
Feature/documentation

lelemmen:
Update the documentation in the source code to work with Doxygen.

### DIFF
--- a/include/FockSpace/BaseFockSpace.hpp
+++ b/include/FockSpace/BaseFockSpace.hpp
@@ -29,10 +29,12 @@ namespace GQCP {
 
 
 /**
- *  A base class for the Fock space
- *  Interfacing requires the Fock space to generate an ONV from a given address
- *  transform a given ONV into the next ONV (in the full or selected space)
- *  and retrieve the address of a given ONV in the space
+ *  A base class for the representation of a Fock space.
+ *
+ *  Derived classes should be able to
+ *      - generate an ONV from a given address
+ *      - transform a given ONV into the 'next' ONV
+ *      - retrieve the address of a given ONV
  */
 class BaseFockSpace {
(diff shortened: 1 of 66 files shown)


include/FockSpace/BaseFockSpace.hpp
```diff
@@ -29,10 +29,12 @@ namespace GQCP {
 
 
 /**
- *  A base class for the Fock space
- *  Interfacing requires the Fock space to generate an ONV from a given address
- *  transform a given ONV into the next ONV (in the full or selected space)
- *  and retrieve the address of a given ONV in the space
+ *  A base class for the representation of a Fock space.
+ *
+ *  Derived classes should be able to
```
tmhuysen: No longer the case. This should be removed.  The Derived classes no longer interface this way to the base class.

include/FockSpace/BaseFockSpace.hpp
```diff
@@ -29,10 +29,12 @@ namespace GQCP {
...
+ *      - generate an ONV from a given address
```
tmhuysen: idem

include/FockSpace/BaseFockSpace.hpp
```diff
@@ -29,10 +29,12 @@ namespace GQCP {
...
+ *      - transform a given ONV into the 'next' ONV
```
tmhuysen: idem

include/FockSpace/BaseFockSpace.hpp
```diff
@@ -29,10 +29,12 @@ namespace GQCP {
...
+ *      - retrieve the address of a given ONV
```
tmhuysen: idem

PR #108, opened 2018-11-02
Fix FindPackages.cmake typo?

matt-chan:
Was this a typo? The CMake docs say it should be COMPONENTS instead if I'm reading them correctly. If you want the components to be optional, you can use OPTIONAL_COMPONENTS. 

https://cmake.org/cmake/help/latest/command/find_package.html


lelemmen:
Yes, I do believe that was a typo! Good catch. I’ll get back to your PRs (also on other libraries) on Monday.

Issue #109, opened 2018-11-05
Tests fail with Intel compilers


lelemmen:
### CMake log

> -- Building gqcp in Release mode
> CMake Warning at /kyukon/home/apps/CO7/haswell-ib/software/CMake/3.10.3-GCCcore-6.4.0/share/cmake-3.10/Modules/FindBoost.cmake:801 (message):
>   New Boost version may have incorrect or missing dependencies and imported
>   targets
> Call Stack (most recent call first):
>   /kyukon/home/apps/CO7/haswell-ib/software/CMake/3.10.3-GCCcore-6.4.0/share/cmake-3.10/Modules/FindBoost.cmake:907 (_Boost_COMPONENT_DEPENDENCIES)
>   /kyukon/home/apps/CO7/haswell-ib/software/CMake/3.10.3-GCCcore-6.4.0/share/cmake-3.10/Modules/FindBoost.cmake:1542 (_Boost_MISSING_DEPENDENCIES)
>   cmake/FindPackages.cmake:5 (find_package)
>   CMakeLists.txt:37 (include)
> 
> 
> -- Boost version: 1.66.0
> -- Found the following Boost libraries:
> --   program_options
> -- cpputil version: 1.5.1
> -- numopt version: 1.5.1
> -- Configuring done
> -- Generating done
> -- Build files have been written to: /user/home/gent/vsc416/vsc41680/GQCG/gqcp/build

### CTest log:
> The following tests FAILED:
> 	  3 - CISolver_FCI_Davidson_test (Failed)
> 	  4 - CISolver_FCI_Dense_test (Failed)
> 	 10 - OO_AP1roG_test (Failed)
> 	 11 - AP1roGPSESolver_test (Failed)
> 	 19 - HamiltonianParameters_test (Failed)
> 	 27 - DIISRHFSCFSolver_test (Failed)
> 	 28 - PlainRHFSCFSolver_test (Failed)
> 	 37 - OO_DOCI_test (Failed)
> 	 38 - RMP2_test (Failed)

lelemmen:
### RMP2 test output


```
Running 2 test cases...
Will read basis set from /apps/gent/CO7/haswell-ib/software/Libint/2.4.2-intel-2018a/share/libint/2.4.2/basis/sto-3g.g94
unknown location(0): fatal error: in "crawdad_sto3g_water": std::runtime_error: The RHF SCF procedure is converged but the MO Fock matrix is not diagonal.
[...]

PR #121, opened 2018-11-09
Add support for standard compilers.

guacke:
This implements #119.


guacke:
This pull request is needed for #109.

CMakeLists.txt
```diff
@@ -19,9 +19,17 @@ set(CMAKE_CXX_STANDARD 11)
 if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE Release)
 endif()
-set(CMAKE_CXX_FLAGS_DEBUG "-g -march=native")
```
lelemmen: Could you keep the ˋ-march=nativeˋ flag for the debug build? When dependencies are installed in release with this flag, this would cause trouble.

cmake/FindPackages.cmake
```diff
@@ -8,7 +8,7 @@ find_package(Boost REQUIRED REQUIRED program_options)
 find_package(Eigen3 3.3.4 REQUIRED)
 
 # Find libint
-find_package(libint2 REQUIRED)
+find_package(Libint2 REQUIRED)
```
lelemmen: I agree with the capitalization: could you do the same for the CMake variables inside ˋFindLibint2.cmakeˋ?

PR #122, opened 2018-11-09
CMake update include and libs.

guacke:
This could be part of the solution to GQCG/gqcp#109.


src/RHF/RHFSCFSolver.cpp
```diff
@@ -81,6 +81,7 @@ void RHFSCFSolver::solve() {
             // After the SCF procedure, we end up with canonical spatial orbitals, i.e. the Fock matrix should be diagonal in this basis
             GQCP::OneElectronOperator F (F_AO);
             F.transform(C);  // transform F to the MO basis with C
+            std::cout << F.get_matrix_representation() << std::endl;
```
lelemmen:
```suggestion
```

PR #133, opened 2018-11-13
Try to speed up Travis compilation.

guacke:
Faster compilations should lead to shorter development iterations. This commit enables make to run in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,77 +13,88 @@ notifications:
     slack: GQCG:5PZ5ibXCwNU1b5ToWbHVIqnI
     email: false
 
+before_cache:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,86 +13,93 @@ notifications:
...
-  # Install a recent version of CMake
-  - conda install -c conda-forge cmake
+# Cache miniconda for faster build procedures
+cache:
+    directories:
+        - $HOME/miniconda
+        - $HOME/download
+        - /tmp/libint
 
+# Before installing, we need a never version of CMake.
```
lelemmen: I would delete this out-dated comment
```suggestion
```

Issue #100, opened 2018-10-31
Implement ONV::reset?


tmhuysen:
https://github.com/GQCG/gqcp/blob/a61b9a82fb364dd6589e1af9ff6fdbe249e05b9a/src/FockSpace/SelectedFockSpace.cpp#L95

Rather than creating a new `ONV` for beta, consider creating the initial ONV outside of the loop, and just resetting it through a reset function `ONV::reset( size_t address)` as an ONV dynamically allocates a small amount of memory.

tmhuysen:
was implemented in #134

PR #134, opened 2018-11-13
can now set an ONV to a desired address

tmhuysen:
This implements issue #100

### DIFF
--- a/include/FockSpace/FockSpace.hpp
+++ b/include/FockSpace/FockSpace.hpp
@@ -105,6 +105,13 @@ public:
      *  @return the address (i.e. the ordering number) of the given ONV
      */
     size_t getAddress(const ONV& onv);
+
+
+    /**
+     *  @param onv          the ONV
+     *  @param address      the address to which the ONV will be set
+     */
+    void setONV(ONV& onv, size_t address);
 };
 
 
--- a/include/FockSpace/FockSpace.hpp
+++ b/include/FockSpace/FockSpace.hpp
@@ -108,10 +108,12 @@ public:
 
 
     /**
+     *  Transform an ONV to one with corresponding to the given address
+     *
      *  @param onv          the ONV
      *  @param address      the address to which the ONV will be set
      */
-    void setONV(ONV& onv, size_t address);
+    void set(ONV &onv, size_t address) const;
 };
 
 
--- a/include/FockSpace/FockSpace.hpp
+++ b/include/FockSpace/FockSpace.hpp
@@ -113,7 +113,7 @@ public:
      *  @param onv          the ONV
      *  @param address      the address to which the ONV will be set
      */
-    void set(ONV &onv, size_t address) const;
+    void set(ONV& onv, size_t address) const;
 };
 
 
--- a/include/FockSpace/ONV.hpp
+++ b/include/FockSpace/ONV.hpp
@@ -22,10 +22,14 @@
 #include <Eigen/Dense>
 
 #include "common.hpp"
+#include <tuple>
 
 
 
 namespace GQCP {
+
+
+using K_N_pair = std::tuple<size_t, size_t>;
 
 
 /**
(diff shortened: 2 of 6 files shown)


include/FockSpace/ONV.hpp
```diff
@@ -22,12 +22,16 @@
...
 
 
 namespace GQCP {
 
 
+using K_N_pair = std::tuple<size_t, size_t>;
+
+
 /**
  *  A class that represents an ONV (occupation number vector).
```
lelemmen:
```suggestion
 *  A class that represents an ONV (occupation number vector)
```

include/FockSpace/ONV.hpp
```diff
@@ -65,6 +69,12 @@ class ONV {
      */
     ONV(size_t K, size_t unsigned_representation);
 
+    /**
+     *  @param K                        the number of orbitals
+     *  @param N                        the number of electrons
+     */
+    ONV(GQCP::K_N_pair pair);
```
lelemmen: I can see why you need a special constructor for this (accepting `K` and `N` would lead to the same constructor declaration as the one with the unsigned representation), but I don't think you should be documenting `K` and `N` in this constructor. Rather, I would skip the `using K_N_pair` and just make the constructor accept a tuple:
```suggestion
    ONV(std::tuple<size_t, size_t> pair);
```
and then document that the pair should be composed of `K` and `N`.

Anyways, maybe some time should be spent on clarifying the API for these constructors?

Maybe we could just delete the `ONV(size_t K, size_t unsigned_representation)` constructor, since we already have one with an extra argument: the number of electrons `N`?

src/FockSpace/ONV.cpp
```diff
@@ -32,28 +32,32 @@ namespace GQCP {
...
 }
 
 /**
  *  @param K                        the number of orbitals
  *  @param unsigned_representation  the representation for the ONV as an unsigned integer
  */
-ONV::ONV(size_t K, size_t representation):
-    ONV(K, __builtin_popcountl(representation), representation)
+ONV::ONV(size_t K, size_t representation) :
+        ONV(K, __builtin_popcountl(representation), representation)
```
lelemmen: remove excessive indentation
```suggestion
    ONV(K, __builtin_popcountl(representation), representation)
```

src/FockSpace/ONV.cpp
```diff
@@ -32,28 +32,32 @@ namespace GQCP {
...
+{}
+
+/**
+ *  @param K                        the number of orbitals
+ *  @param N                        the number of electrons
+ */
+ONV::ONV(K_N_pair pair) :
+        K(std::get<0>(pair)),
```
lelemmen:
```suggestion
    K(std::get<0>(pair)),
```

src/FockSpace/ONV.cpp
```diff
@@ -32,28 +32,32 @@ namespace GQCP {
...
+        N(std::get<1>(pair))
```
lelemmen:
```suggestion
    N(std::get<1>(pair))
```

src/FockSpace/ONV.cpp
```diff
@@ -32,28 +32,32 @@ namespace GQCP {
...
 {
     occupation_indices = VectorXs::Zero(N);
```
lelemmen:
```suggestion
    this->occupation_indices = VectorXs::Zero(N);
```

src/FockSpace/ONV.cpp
```diff
@@ -32,28 +32,32 @@ namespace GQCP {
...
-    ONV(K, __builtin_popcountl(representation), representation)
+ONV::ONV(size_t K, size_t representation) :
+        ONV(K, __builtin_popcountl(representation), representation)
+{}
+
+/**
+ *  @param K                        the number of orbitals
+ *  @param N                        the number of electrons
+ */
+ONV::ONV(K_N_pair pair) :
```
lelemmen: Could you also document which ONV is made with this constructor? (Every ONV inside the F(K,N) Fock space has the same number of orbitals and electrons)

include/FockSpace/FockSpace.hpp
```diff
@@ -108,10 +108,12 @@ class FockSpace: public GQCP::BaseFockSpace {
 
 
     /**
+     *  Transform an ONV to one with corresponding to the given address
+     *
      *  @param onv          the ONV
      *  @param address      the address to which the ONV will be set
      */
-    void setONV(ONV& onv, size_t address);
+    void set(ONV &onv, size_t address) const;
```
lelemmen: `set(ONV& onv, size_t address)`

PR #141, opened 2018-11-20
Feature/scalar

lelemmen:


lelemmen:
@GQCG/cpp-developers I'm wondering if we should change `calculateMolecularHamiltonianParameters()` such that it also contains the nuclear interaction term. 

Or maybe add another function `calculateMolecularHamiltonianParameters(...,molecule)` which then sets the scalar interaction term based upon `molecule.calculateInternuclearRepulsionEnergy()`?

lelemmen:
I don't think that's a good idea, because different 'molecules' (think ions) can have the same AO basis. I could just implement `calculateMolecularHamiltonianParameters(molecule, basisset_name)`, then the user doesn't even have to create an `AOBasis`: this would then be done inside this function.

lelemmen:
See #156 too.

Issue #110, opened 2018-11-05
Add logging options


lelemmen:
See [boost.log](https://www.boost.org/doc/libs/1_68_0/libs/log/doc/html/index.html)

tmhuysen:
Redirect libint terminal output to boost logs.

lelemmen:
Add the git SHA too.

tmhuysen:
I will start with an initial results oriented log for the reproduction of my thesis results.

tmhuysen:
Discuss levels.
Backend possibilities:
-HDF5
-JSON

lelemmen:
I'm closing this issue because it is very old. It dates back from the time we didn't have Python bindings.

PR #164, opened 2018-11-30
removed constrained_DOCI from cmake

tmhuysen:


lelemmen:
We'll get back to this after the logging PR #110.

PR #165, opened 2018-11-30
Feature/geminal expansion

lelemmen:
In this branch, I'll implement how to get seniority-zero expansion coefficients from AP1roG- and APIG geminal coefficients.


tests/geminals/APIGGeminalCoefficients_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+    size_t N_P = 1;
+
+    Eigen::VectorXd g (3);
+    g << 1, 2, 3;
+    GQCP::APIGGeminalCoefficients gem_coeff (g, N_P, K);
+
+    Eigen::VectorXd ref_coefficients (3);
+    ref_coefficients << 1, 2, 3;
+
+    GQCP::FockSpace fock_space (K, N_P);
```
tmhuysen: You could add a function to `BaseAPIGGeminalCoefficients" that returns a FockSpace, rather having to re-enter the your constructor arguments in both FockSpace and Geminal class.

lelemmen: That's a good suggestion. However, I feel like this would be just a workaround since we don't have the `toWaveFunction()`-issue cleared up yet.

PR #173, opened 2018-12-06
Bug fix: cluster compilation.

guacke:
In this issue, I will resolve all current cluster compilation problems.


guacke:
This commit turns benchmarks off by default. Please pass "-DBUILD_BENCHMARKS=ON" to enable benchmarks.

PR #178, opened 2018-12-07
Install executables in binary directory.

guacke:
This commit causes the executables to be installed in the ${INSTALL_PREFIX}/bin directory, instead of remaining in the build directory.


guacke:
I have added this as feature request #180. I propose to address this feature in a different PR.

lelemmen:
@guacke Can't we just do it in this one?

lelemmen:
The previous commit implements #180.

Issue #153, opened 2018-11-23
Shared libraries for gqcp?


lelemmen:
Is there a possibility to build several dynamic libraries for the gqcp?

guacke:
As the move to the shared library will be a concerted effort among the libraries, I will report here when all feature/shared_library are ready to be merged to develop.

lelemmen:
After resetting the caches, we can compile the dynamic library on Travis again, after the cpputil-numopt-gqcp merger (#184).

PR #181, opened 2018-12-07
Move to shared libraries. 

guacke:
Revert travis.yml to develop branches before merging with develop.


guacke:
This commit is needed for GQCG/gqcp#153.

Issue #166, opened 2018-12-03
Keep cpputil and numopt?


lelemmen:
Should we keep the separate cpputil and numopt libraries? Currently, for us, they're only used in this project and not in others.

lelemmen:
cpputil and numopt were merged into this library in #184.

PR #184, opened 2018-12-13
Feature/merger

lelemmen:
In this PR, cpputil and numopt are included into the main library. See #166.


lelemmen:
That's weird: the boost-libint bug seems to be back again.

lelemmen:
If we encounter any bugs that have been introduced by merging cpputil and numopt in this library, I'm sorry, but I think this was the cleanest and fastest way to get an up-to-date develop branch.

Issue #185, opened 2018-12-14
Libint-boost


lelemmen:
As of #184, it seems that the libint-boost (see also #96) has resurfaced on Travis. On my machine, there isn't anything wrong, though.

PR #194, opened 2018-12-14
Fix libint-boost-related bug.

guacke:
This pull request is aimed at resolving issue #185.


lelemmen:
As discussed, the libint-boost bug seems to be related to the order of the includes in the tests: libint shouldn't be included before any boost library.

Issue #42, opened 2018-10-15
Const correctness


tmhuysen:
Consider making the code const correct.

CISolver, FockSpace, WaveFunction (also see numopt)

lelemmen:
`FockSpace` has a const dimension, while `SelectedFockSpace` doesn't.

lelemmen:
This was implemented in #199.

Issue #171, opened 2018-12-04
Remove `GQCP::` in code that is wrapped inside the namespace


lelemmen:
If code is wrapped inside `namespace GQCP { ... }`, there is no need to specify extra `GQCP::` namespace identifiers.

lelemmen:
For classes, I would definitely remove the `GQCP::` namespace, but for separate functions, such as `GQCP::areEqualEigenvectors(...)`, I think the namespace should be included, just for readability.

lelemmen:
This was implemented in #199.

Issue #169, opened 2018-12-04
Rename FockSpace 'get_ONV'


tmhuysen:
Method at : https://github.com/GQCG/gqcp/blob/0c8039a0d0479a5bbf35182f7aa500619b006124/include/FockSpace/FockSpace.hpp#L92

should not be named `get_ONV(size_t address)` as it generates the ONV rather then return a reference to an existing one. `makeONV(size_t address)` for example would be more suited.

lelemmen:
This is related to #123.

lelemmen:
This was implemented in #199.

Issue #161, opened 2018-11-30
Rename `get_occupied_index`


lelemmen:
For the `ONV` class, we already have the getter `get_occupation_indices()`. Unfortunately, the getter that corresponds to just getting the occupation index of 1 electron instead of them all is named `get_occupied_index()`.

I suggest renaming `get_occupied_index()` to `get_occupation_index()` to maintain consistency.

lelemmen:
This was implemented in #199.

PR #199, opened 2018-12-14
Feature/const

lelemmen:
This PR implements the const-correct feature we've been dreaming about. See #42.

### DIFF
--- a/include/optimization/BaseEigenproblemSolver.hpp
+++ b/include/optimization/BaseEigenproblemSolver.hpp
@@ -34,11 +34,11 @@ namespace GQCP {
  */
 class BaseEigenproblemSolver {
 protected:
-    const size_t dim;  // the dimension of the vector space associated to the eigenvalue problem
-    const size_t number_of_requested_eigenpairs;
+    size_t dim;  // the dimension of the vector space associated to the eigenvalue problem
+    size_t number_of_requested_eigenpairs;
 
     bool _is_solved = false;
-    std::vector<const Eigenpair> eigenpairs;  // a collection of the eigenpairs of the eigenproblem, sorted with increasing eigenvalue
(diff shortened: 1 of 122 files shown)


lelemmen:
It also implements #161, #169, #171 and #147.

tmhuysen:
Some of the optimization tests are failing at random, some of the times.

Issues appear to be minor accuracy changes.
eg. 0.999999999996 for the norm of an eigenvector. This fails the 10e-12 threshold (would succeed for 10e-11)

I've been testing a local old numopt code with the same parameters:
>gqcp compiled "-g -march=native, EIGEN3 3.3.5"
>numopt compiled "-g -march=native, EIGEN3 3.3.5"

The test that is ran : "liu_50_collapse"

The gqcp code fails approximately 1/2 of the times.
The numopt code does not fail.

I've been gradually changing numopt code to that of current live GQCP version. No fails so far.

include/optimization/BaseEigenproblemSolver.hpp
```diff
@@ -59,50 +55,30 @@ class BaseEigenproblemSolver {
...
     // GETTERS
-    /**
-     *  @return the diagonal of the matrix representation of the matrix whose eigenproblem is being solved
-     */
-    virtual Eigen::VectorXd get_diagonal() = 0;
     bool is_solved() const { return this->_is_solved; }
 
-
-    // GETTERS - EIGENPAIR
     std::vector<Eigenpair> get_eigenpairs() const;
```
tmhuysen: This line should return a const reference not a copy, this causes the sudden drop in accuracy.

lelemmen: Sorry, my bad. Thanks for catching this!

PR #204, opened 2018-12-18
Feature/fockspace_iterator

lelemmen:
In this PR, we will implement iterators over
* [x] `FockSpace`
* [x] `SelectedFockSpace`,

implementing #206.


lelemmen:
@tmhuysen The main benefit of using the iterator classes was to clean up the iteration, so that you don't have to do things like

```cpp
        // Skip the last permutation
        if (I < dim-1) {
            this->fock_space.setNextONV(onv);
        }
```

PR #205, opened 2018-12-18
Make finding data for tests more robust.

guacke:
The HPC EasyBuild framework builds the libraries 'completely' out-of-source, meaning, in a directory above the top source folder. This commit makes sure that the test data can also be found in that case.


tests/Molecule_test.cpp
```diff
@@ -96,17 +96,17 @@ BOOST_AUTO_TEST_CASE ( parseXYZFile ) {
     BOOST_REQUIRE_THROW(GQCP::Molecule::Readxyz("this is a nonsense data path"), std::runtime_error);
 
     // Make sure we get an error when a path with a wrong extension is given
-    BOOST_REQUIRE_THROW(GQCP::Molecule::Readxyz("../tests/ref_data/nuclear.data"), std::runtime_error);
+    BOOST_REQUIRE_THROW(GQCP::Molecule::Readxyz("ref_data/nuclear.data"), std::runtime_error);
```
lelemmen: I see now that I made a mistake in a previous commit. This should be `data/small_vector.data`

PR #207, opened 2018-12-19
Feature/rdm calculator refactor

tmhuysen:


include/RDM/UnresolvedCIRDMBuilder.hpp
```diff
@@ -0,0 +1,81 @@
...
+    BaseFockSpace* get_fock_space() override { return &fock_space; }
+
+
+    // OVERRIDDEN PUBLIC METHODS
+    /**
+     *  @param x        the coefficient vector representing the UnresolvedCI wave function
+     *
+     *  @return all 1-RDMs given a coefficient vector
+     */
+    OneRDMs calculate1RDMs(const Eigen::VectorXd& x) const override;
```
lelemmen: Unfortunately, since this is a `SpinUnresolved`-type `RDMBuilder`,  this should return a `OneRDM` instead of `OneRDMs`.

I think this issue can be avoided if we insert two other base classes: `BaseSpinResolvedRDMBuilder` and `BaseSpinUnresolvedRDMBuilder`, that inherit from BaseRDMBuilder (since they all can provide a `calculateElement`) and define:
* `virtual OneRDM calculate1RDM() const = 0` for unresolved
* `virtual OneRDMs calculate1RDMs() const = 0` for resolved

src/DOCINewtonOrbitalOptimizer.cpp
```diff
@@ -76,19 +76,18 @@ const Eigenpair& DOCINewtonOrbitalOptimizer::get_eigenpair(size_t index) const {
 void DOCINewtonOrbitalOptimizer::solve(BaseSolverOptions& solver_options, const OrbitalOptimizationOptions& oo_options) {
     this->is_converged = false;
     auto K = this->ham_par.get_K();
-
+    RDMCalculator rdm_calculator(*this->doci.get_fock_space());
```
lelemmen:
```suggestion
     RDMCalculator rdm_calculator(*this->doci.get_fock_space());  // make the RDMCalculator beforehand, it doesn't have to be constructed in every iteration
```

tests/RDM/RDMCalculator_test.cpp
```diff
@@ -55,7 +55,70 @@ BOOST_AUTO_TEST_CASE ( constructor ) {
...
+    // Create a test wave function
+    size_t M = 3;
+    size_t N = 1;
+    GQCP::FockSpace fock_space (M, N);
+
+    Eigen::VectorXd coeff (fock_space.get_dimension());
+    coeff << 1, 2, -3;
+    GQCP::RDMCalculator d = GQCP::RDMCalculator::SpinUnresolved(fock_space);
+    d.set_coefficients(coeff);
+    BOOST_CHECK_THROW(d(0), std::invalid_argument);
```
lelemmen:
```suggestion
     BOOST_CHECK_THROW(d(0), std::invalid_argument);  // need an even number of indices
```

include/RDM/UnresolvedCIRDMBuilder.hpp
```diff
@@ -0,0 +1,81 @@
...
+/**
+ *  A class capable of calculating 1- and 2-RDMs from wave functions expanded in the full CI product Fock space
+ */
+class UnresolvedCIRDMBuilder : public BaseRDMBuilder {
+    FockSpace fock_space;  // Fock space containing the alpha and beta Fock space
+
+
+public:
+    // CONSTRUCTORS
+    explicit UnresolvedCIRDMBuilder(const FockSpace& fock_space);
```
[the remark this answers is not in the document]
tmhuysen: I would keep SpinResolved as default, and Unresolved as a special case. In general SpinResolved FCI is referenced to as FCI.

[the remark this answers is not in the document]
lelemmen: SpinResolved as default is fine.

PR #212, opened 2018-12-21
Bug fix: do not use separate 'gqcp' install directory.

guacke:
We should not use the additional 'gqcp' directory inside installation directories, as this complicates finding the executables and libraries (i.e. manually adjusting paths).


lelemmen:
We have discussed that using an installation pattern like `gqcp/{include, lib, cmake, bin}`, as we currently have, is less prone to compilation errors, since our CMake configuration takes care of the include paths and targets through `gqcp_INCLUDE_DIRS` and the `gqcp` library target.

On the other hand, this forces our users to use a CMake-approach, but we feel that it is good practice to adopt more modern developing tools.

PR #245, opened 2019-02-25
public functions for executables

tmhuysen:


tmhuysen:
The features of this PR will be addressed in a different PR.

Issue #240, opened 2019-02-15
Syntax improvements


lelemmen:
From #237, we are now able to do

```cpp
OneElectronOperator<double> T, V;
auto H = OneElectronOperator<double>(T + V);
```

without implementing our own `operator+()` for `OneElectronOperator`s. Can we improve the implementation such that we can use the syntax

```cpp
auto H = T + V;
```
?

lelemmen:
Maybe we could use: https://en.cppreference.com/w/cpp/language/cast_operator?

Issue #96, opened 2018-10-28
Too many include directories for libint?


lelemmen:
I was thinking about the boost/libint bug we had, and maybe we are including one too many directories for libint? The libint documentation specifies that we only need to `#include <libint2.hpp>` which is in e.g. `/usr/local/libint/2.3.1/include`.

Maybe we don't need `${LIBINT_PREFIX}/include/libint2` in the following line?

https://github.com/GQCG/gqcp/blob/0c4ccf1457b842b495c92164f8eb8aa8bf390b51/cmake/Findlibint2.cmake#L24

PR #251, opened 2019-03-03
Implement Matrix and Tensor classes

lelemmen:
In this PR, we will implement `GQCP::Matrix` and `GQCP::Tensor` as derived classes from their `Eigen` counterparts, such that we can extend these base classes and inherit from them in for example `SquareMatrix`.


lelemmen:
As we can see in the tests, we have succeeded in creating the API required from issue #240.

lelemmen:
Some information about the libint-boost bug in #96. If the preprocessor finds libint/boost/preprocessor first, then the macros of the family `BOOST_PP_ITERATE` are set using libint paths. If we then use another part of boost's library, like boost/string, that also uses `BOOST_PP_ITERATE`, these macros are set to the wrong directories.

I've managed to fix this bug by including `<boost/preprocessor.hpp>` before including `<libint2.hpp>`, so libint's own boost is ignored. I hope this doesn't cause any problems on other machines.

include/math/SquareRankFourTensor.hpp
```diff
@@ -0,0 +1,271 @@
...
+     *  Example:
+     *      Given a rank-4 tensor of dimensions (10,10,10,10), and a matrix of dimensions (3,3)
+     *      When choosing r and s as 2,1 respectively, and i,j,k,l as 1,2,1,0
+     *      The matrix is added to the tensor with 1st index starting at i and 4th index starting at l held fixed at 1 and 0 respectively
+     *      While 2nd index starting from j (2) now is mapped to the 2nd matrix (s=1) element and 3rd index starting from k (1) is mapped to 1st (r=2)
+     *      meaning that the additions follow : tensor(1,2,1,0) += M(0,0), tensor(1,3,1,0) += M(0,1), ..., tensor(1,3,4,0) += M(3,1), tensor(1,4,4,0) += M(3,2), etc
+     *      and for other indices we find : tensor(2,2,1,0) += 0 (no changes, these are obviously not performed)
+     */
+    template<size_t r, size_t s>
+    Self& addBlock(const MatrixX<Scalar>& M, size_t i, size_t j, size_t k, size_t l) {
```
tmhuysen: This method is not exclusive to a square tensor.

include/math/Tensor.hpp
```diff
@@ -0,0 +1,245 @@
...
+
+
+    /*
+     *  PUBLIC METHODS
+     */
+
+    /**
+     *  @return this as an Eigen::Tensor, as a work-around to fix Eigen::Tensor expressions
+     */
+    const Base& Eigen() const {
```
tmhuysen: Maybe we can find a better solution to this in the future

include/math/SquareRankFourTensor.hpp
lelemmen: Indeed, I missed that one. Done.

include/math/Tensor.hpp
lelemmen: Apparently `Eigen/Dense` did something so that we can do operations with derived classes of `Eigen::Matrix`, but that kind of support is not yet included in `Eigen/Tensor`. I hope that one day `Eigen/Tensor` does the same thing.

lelemmen:
> I also don't quite get :
> 
> > * @tparam ExpDerived      the type of the Eigen expression (normally generated by the compiler)
> > * @tparam AccessLevel     the access level of the TensorBase (normally generated by the compiler)


If you look at [Tensor.hpp](https://eigen.tuxfamily.org/dox/unsupported/Tensor_8h_source.html), line 384, we can see a constructor for `ReadOnlyAccessors`

```cpp
template<typename OtherDerived>
EIGEN_DEVICE_FUNC
EIGEN_STRONG_INLINE Tensor(const TensorBase<OtherDerived, ReadOnlyAccessors>& other) 
{ ... }
```

and afterwards one for `WriteOnlyAccessors`. That is the constructor I 'copied' for our derived `GQCP::Tensor` class, so we can handle both `AccessLevel==WriteOnlyAccessors` and `AccessLevel==ReadOnlyAccessors`. The name `AccessLevel` for this template argument is one that I took from the [`Eigen::TensorBase` documentation](https://eigen.tuxfamily.org/dox/unsupported/classEigen_1_1TensorBase.html).

> But do not feel like these should be elaborated upon in the code itself, it's clear it's something to confirm with the Eigen library

I hope this answer is sufficiently clear, but as far as I know it's something that is necessary to derive from `Eigen::Tensor`. Unfortunately deriving from `Eigen::Tensor` is undocumented, which is why I had to do the kind of band-aid fix (that you rightfully commented on) for `GQCP::Tensor`:

```cpp
const Base& Eigen() const {
    return static_cast<const Base&>(*this);
}
```

PR #257, opened 2019-03-07
Make the DIIS algorithm more flexible

lelemmen:
I did a force push to reset the changes that had been made, as I think updating the DIIS algorithm doesn't have to include the conditioning number step, since that seems to have no impact whatsoever.

The DIIS algorithm will be more flexible if we
- keep the `maximum_subspace_dimension`, such that if there are too many Fock matrices, the oldest one is discarded 
- introduce a `minimum_subspace_dimension`, such that DIIS is only enabled when there have been at least this many iterations (and thus Fock matrices)

### DIFF
--- a/include/RHF/DIISRHFSCFSolver.hpp
+++ b/include/RHF/DIISRHFSCFSolver.hpp
@@ -32,10 +32,11 @@ namespace GQCP {
 
 class DIISRHFSCFSolver : public RHFSCFSolver {
 private:
+    size_t minimum_subspace_dimension;
     size_t maximum_subspace_dimension;
 
     std::deque<OneElectronOperator<double>> fock_matrix_deque = {};  // deque of Fock matrices used in the DIIS algorithm
-    std::deque<SquareMatrix<double>> error_matrix_deque = {};  // deque of error matrices used in the DIIS algorithm
+    std::deque<OneElectronOperator<double>> error_matrix_deque = {};  // deque of error matrices used in the DIIS algorithm
 
 
     // PRIVATE METHODS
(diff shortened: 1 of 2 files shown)


include/RHF/DIISRHFSCFSolver.hpp
```diff
@@ -32,10 +32,11 @@ namespace GQCP {
 
 class DIISRHFSCFSolver : public RHFSCFSolver {
 private:
+    size_t minimum_subspace_dimension;
```
tmhuysen: Add the descriptions from the constructor to the member variables:
 >*  @param minimum_subspace_dimension       the minimum number of Fock matrices that have to be in the subspace before enabling DIIS
     *  @param maximum_subspace_dimension       the maximum DIIS subspace dimension before the oldest Fock matrices get discarded (one at a time)

src/RHF/DIISRHFSCFSolver.cpp
```diff
@@ -75,10 +76,16 @@ OneElectronOperator<double> DIISRHFSCFSolver::calculateNewFockMatrix(const OneRD
             f_AO += y(i) * this->fock_matrix_deque[i];
         }
 
-        // Remove the oldest entries, which means that we collapse every iteration once the dimension is large enough
+
+    }  // subspace collapse
+
+
+    // Collapse the subspace if it becomes too large, discard the oldest entry
```
tmhuysen: is this considered a "collapse"?

lelemmen:
Thanks for the review. I've updated the documentation.

Issue #249, opened 2019-03-01
Error throws should better indicate their origin


tmhuysen:
As has been done in recent PRs, all error throws should better indicate their origin in the code when thrown.

PR #258, opened 2019-03-11
Feature/error locations

tmhuysen:
Adds function names to errors it throws. This implements #249


tmhuysen:
This PR resolves #258

tmhuysen:
> I also saw that in some cases you have `const &` and in others you don't: is there a specific reason?

No, this should be consistent in the next commit, but since you cannot overload function with  `(const instance& name)` vs `(instance name)` leaving that part out is not ambiguous and decided on the latter.

Issue #267, opened 2019-04-29
I cannot compile my separate drivers  on the cluster.


tmhuysen:
>In file included from /user/gent/425/vsc42536/drivers/swalot/tmhuysen-gqcp-drivers/drivers/test_driver.cpp(5):
/data/gent/425/vsc42536/apps/swalot/gqcg/gqcp/include/gqcp.hpp(63): catastrophic error: cannot open source file "Operator/BaseOperator.hpp"
  #include "Operator/BaseOperator.hpp"

BaseOperator.hpp does no longer exist but is included in the gqcp.hpp

tmhuysen:
Has been resolved.

Issue #266, opened 2019-04-29
Develop does not compile on the cluster


tmhuysen:
>/scratch/gent/425/vsc42536/apps/swalot/gqcg/gqcp/src/Basis/LibintInterfacer.cpp(242): error: no default constructor exists for class "GQCP::ShellSet"
      ShellSet shell_set;

>compilation aborted for /scratch/gent/425/vsc42536/apps/swalot/gqcg/gqcp/src/Basis/LibintInterfacer.cpp (code 2)
make[2]: *** [src/CMakeFiles/gqcp.dir/Basis/LibintInterfacer.cpp.o] Error 2

>make[1]: *** [src/CMakeFiles/gqcp.dir/all] Error 2
make: *** [all] Error 2

Cluster: Swalot
Compiler flags: -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_COMPILER=icpc

I appears the intel compiler is too strict regarding default initialization of `ShellSet`. Even though it inherits base constructors from `std::vector<Shell>` .

tmhuysen:
has been resolved

PR #268, opened 2019-04-29
Feature/intel compiler conformity

tmhuysen:
PR fixes requirements for the intel compiler on the cluster.  Also updated the gqcp.hpp header (which only appeared to be an issue on the cluster, not for local compilation with clang)

This PR addresses:
#267 
#266

### DIFF
--- a/include/Basis/ShellSet.hpp
+++ b/include/Basis/ShellSet.hpp
@@ -38,6 +38,7 @@ public:
 
 public:
     // CONSTRUCTORS
+    ShellSet() = default;
     /**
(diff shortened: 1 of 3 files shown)


include/Basis/ShellSet.hpp
```diff
@@ -38,6 +38,7 @@ class ShellSet : public std::vector<Shell> {
 
 public:
     // CONSTRUCTORS
+    ShellSet() = default;  // required for the intel compiler on the cluster
```
lelemmen:
```suggestion
    ShellSet() = default;  // required for Intel compilers
```

PR #275, opened 2019-05-21
CMake update: targets and properties.

guacke:
This pull request will make the CMake build more modern, adhering to the philosophy of targets and properties (https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/).

### DIFF
--- a/cmake/Find/FindLibcint.cmake
+++ b/cmake/Find/FindLibcint.cmake
@@ -12,6 +12,7 @@
 
 
 # When the user has set ${LIBCINTROOT} or ${LIBCINT_ROOT} in the enviroment, this path can also be used
+
 find_path(LIBCINT_PREFIX include/cint.h HINTS /usr/local/*/ $ENV{LIBCINTROOT} $ENV{LIBCINT_ROOT} ${LIBCINTROOT} ${LIBCINT_ROOT})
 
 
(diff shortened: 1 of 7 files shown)


guacke:
Travis now uses Conda to install libcint.

cmake/Find/FindLibcint.cmake
```diff
@@ -27,14 +28,24 @@ else()
...
     if(EXISTS ${LIBCINT_PREFIX}/lib/libcint.so)
-        set(Libcint_LIBRARIES "${Libcint_LIBRARIES};${LIBCINT_PREFIX}/lib/libcint.so")
+        set(Libcint_LIBRARIES "${LIBCINT_PREFIX}/lib/libcint.so")
     else()
         if(EXISTS ${LIBCINT_PREFIX}/lib/libcint.dylib)
-            set(Libcint_LIBRARIES "${Libcint_LIBRARIES};${LIBCINT_PREFIX}/lib/libcint.dylib")
+            set(Libcint_LIBRARIES "${LIBCINT_PREFIX}/lib/libcint.dylib")
         else()
-            set(Libcint_LIBRARIES "${Libcint_LIBRARIES};${LIBCINT_PREFIX}/lib/libcint.a")
+            set(Libcint_LIBRARIES "${LIBCINT_PREFIX}/lib/libcint.a")
```
lelemmen: I think I missed an if-case during my last PR

```cmake
if(EXISTS ${LIBCINT_PREFIX}/lib/libcint.a)
```

Otherwise, `libcint.a` gets set even though it may not be present.

cmake/Find/FindLibint2.cmake
```diff
@@ -38,3 +38,12 @@ else()
```
lelemmen: Idem FindLibcint.cmake. Could you fix it here as well?

cmake/Find/FindLibcint.cmake
lelemmen: I think I missed an if-case during my last PR

```cmake
if(EXISTS ${LIBCINT_PREFIX}/lib/libcint.a)
```

Otherwise, `libcint.a` gets set even though it may not be present.

PR #291, opened 2019-06-05
Feature/cluster mkl

guacke:
This PR will fix the failing tests on the cluster that seem to be related to MKL (on the cluster) transposing the eigenvectors of the GeneralizedSelfAdjointEigenSolver.


guacke:
This commit caused **all** test of OO_DOCI_test to fail (instead of only two out of four. As such, the problem we have is not related to any transposition of the eigenvectors.

PR #297, opened 2019-06-06
added 1 and 2 rdms to the standard out

tmhuysen:


lelemmen:
@nbilliet Should the output have some keywords that clarify what is being printed? Or is bare output enough, like proposed in the PR?

PR #301, opened 2019-06-13
Implement OO-pCCD

lelemmen:
In this PR, I will
- [x] implement OO-pCCD


lelemmen:
Unfortunately, due to the ill-implemented Newton-based orbital optimization algorithm in `NewtonOrbitalOptimizer`, the norm of the gradient is not converging, even though the energy is.

The issue is with the Newton step: even though it may be ill-defined (for indefinite Hessians), I am still using it as a valid new step. I will now 
- [ ] implement a simple Hessian modification (Algorithm 3.3 in Nocedal and Wright) to fix this issue.

lelemmen:
Apparently, what I've been trying to do with the `Options`, is just the builder pattern. I'm removing the `Options` classes and postponing the builder pattern to be fixed in a later PR.

PR #313, opened 2019-06-21
Implement electric linear response equations

lelemmen:
In this PR, I will implement the electric linear response equations for:

- [x] RHF
- [x] CI
- [x] AP1roG/pCCD

Furthermore, I will also implement the calculation of the electric polarizability for

- [x] Rayleigh-Ritz energy functions (RHF and CI)
- [x] Lagrangian 'energy' functions (AP1roG/pCCD)


lelemmen:
There are a couple of things I've also done in this branch, things that I encountered during the implementation. 

Mainly, I've implemented two `Block` classes, `BlockMatrix` and `BlockRankFourTensor`, that behave like the mathematical quantities `G_i^a` (geminal coefficients like in the notes) and `J_{ia, jb}` (elements of the Jacobian of the PSEs). However, I'm not sure about the naming: I think we can decide on this when we're doing the next representation refactor (as we have alluded to in #314). Some things are hanging, for example `APIGVariables` is definitely related to `BlockMatrix`.

Furthermore, I've implemented `AP1roGPSEs` that represents the AP1roG PSEs in a way that I think is an example of an implementation of a base class of `NonLinearEquations` that should be looked into in the next solver refactor.

lelemmen:
Something went wrong during the rebase: the renaming to the capital directories was a big pain.

lelemmen:
I've closed #313 because too many things were done at the same time and I've lost oversight. I will be splitting it over several PRs. I've opened the following issues: #321, #322, #323, #324, #325, #328.

PR #315, opened 2019-06-23
Feature/python (Closes #306)

### LINKED ISSUE #306: Python interface
It would be great feature to have a python interface, instead of a wide array of drivers and cumbersome input files.

guacke:


guacke:
Switched previous PR to draft PR.

lelemmen:
Since this PR will be very large, could we try to split it up in several PRs? I think we should mainly establish which Python bindings are necessary to have right now (like integrals), and establish a roadmap for the other, less important, Python bindings.

PR #317, opened 2019-06-24
Bugfix/cmake (Closes #316).

### LINKED ISSUE #316: I cannot link GQCP to my drivers
Using `find_package(gqcp 0.2.0 NO_MODULE) `
results in error:
>CMake Error at /usr/local/cmake/gqcp-config.cmake:27 (find_dependency):
  Unknown CMake command "find_dependency".

guacke:


guacke:
I can compile drivers with:
```
cmake .. -DCMAKE_PREFIX_PATH="~/local;/anaconda3/envs/gqcg" -DCMAKE_MODULE_PATH="~/local/cmake"
```
I am still looking for a way how we can avoid setting the `MODULE_PATH` explicitly.

guacke:
Related discussion:
https://stackoverflow.com/questions/20746936/cmake-of-what-use-is-find-package-if-you-need-to-specify-cmake-module-path-an

So, my guess is that as long as the depending libraries do not expert their libraries as targets, we will have to include `-DCMAKE_MODULE_PATH` as we are providing these targets ourselves. This seems the better option than:
1. Force installing our Find*.cmake scripts in default CMake paths (can overwrite user defined defaults).
2. Removing the dependencies from the gqcg-config (destroys portability of compiled libraries because we do not explicitly check if dynamic libraries are available to drivers).

PR #334, opened 2019-06-29
Add support for Python bindings (closes #306, closes #340)

### LINKED ISSUE #306: Python interface

### LINKED ISSUE #340: Refactor Python Bindings
As suggested by @lelemmen:

It seems that a lot of things are happening inside the constructor of the HubbardDriver. In my opinion, it would be better to only use the constructor to construct the valid context of the object, and then use a method like solve() or so to do the actual calculations and ‘set’ the energies/DMs that the user would like to use.

I’ve added some comments and suggestions.

As far as the structuring of the Python binding goes, what do you think of the following idea? We could include a folder gqcpy (or pygqcp, I think we should still vote on that because both of them are fine for me) and then structure the python bindings as in the C++ source code itself. For the current Hubbard driver, this would then mean to create the Python module in : root/pygqcp/Drivers/.

guacke:


include/Drivers/HubbardDriver.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include <boost/algorithm/string.hpp>
+#include <boost/program_options.hpp>
+
+#include "HamiltonianParameters/HamiltonianParameters.hpp"
+#include "CISolver/CISolver.hpp"
+#include "HamiltonianBuilder/Hubbard.hpp"
+#include "RDM/RDMCalculator.hpp"
+
+namespace GQCP {
+    class HubbardDriver {
```
lelemmen: Add a class description

lelemmen: No indentation inside a namespace needed

src/Drivers/HubbardDriver.cpp
```diff
@@ -0,0 +1,59 @@
...
+        std::vector<std::string> splitted_line;
+        boost::split(splitted_line, triagonal_line, boost::is_any_of(","));
+
+        std::vector<double> triagonal_data;
+        for (const std::string &x : splitted_line) {
+            triagonal_data.push_back(std::stod(x));
+        }
+
+        triagonal = Eigen::Map<Eigen::VectorXd>(triagonal_data.data(), triagonal_data.size());
+        GQCP::HoppingMatrix H = GQCP::HoppingMatrix::FromUpperTriangle(triagonal);
```
lelemmen: I think it would be useful if the previous lines of code that produces a `HoppingMatrix` from a `std::string` was actually made into a named constructor inside the `HoppingMatrix` class.

src/Drivers/HubbardDriver.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+        // Actual calculations
+        auto ham_par = GQCP::HamiltonianParameters<double>::Hubbard(H);
+        if (ham_par.get_K() != K) {
```
lelemmen: Can’t `K` be determined from the given input `csline`?

src/Drivers/HubbardDriver.cpp
```diff
@@ -0,0 +1,59 @@
...
+            throw std::invalid_argument("hubbard(driver): The given number of sites does not match the triagonal");
+        }
+
+
+        // Initialize and solve the Hubbard eigenvalue problem
```
lelemmen: Since only a `solve` method would use the `ProductFockSpace`, `Hubbard` Hamiltonian builder and `CISolver`, I don’t think it is necessary to include these as members of the class (they don’t need to exist outside of the actual calculation part).

Removing these objects as members from `HubbardDriver` would bypass the issue that you tried to solve by using `shared_ptr`s.

src/Drivers/HubbardDriver.cpp
```diff
@@ -0,0 +1,59 @@
...
+        this->fock_space = std::make_shared<GQCP::ProductFockSpace>(K, num_alpha, num_beta);
+        this->hubbard = std::make_shared<GQCP::Hubbard>(*(this->fock_space));
+        this->solver = std::make_shared<GQCP::CISolver>(*(this->hubbard), ham_par);
+        this->dense_solver_options.number_of_requested_eigenpairs = num_states;
+        this->solver->solve(this->dense_solver_options);
+    }
+
+    std::vector<double> GQCP::HubbardDriver::get_energies() {
+        std::vector<double> energies;
```
lelemmen: In the `solve` method, you could calculate and ‘store’ the targeted energies and then just return them here.

src/Drivers/HubbardDriver.cpp
```diff
@@ -0,0 +1,59 @@
...
+        for (const GQCP::Eigenpair &eigenpair : this->solver->get_eigenpairs()) {
+            energies.push_back(eigenpair.get_eigenvalue());
+        }
+        return energies;
+    }
+
+    std::vector<Eigen::MatrixXd> GQCP::HubbardDriver::get_first_order_rdms() {
+        std::vector<Eigen::MatrixXd> rdms;
```
lelemmen: Idem for the DMs.

src/Drivers/HubbardDriver.cpp
```diff
@@ -0,0 +1,59 @@
...
+        for (const GQCP::Eigenpair &eigenpair : this->solver->get_eigenpairs()) {
+            GQCP::RDMCalculator rdm_calculator(*(this->fock_space));
+            rdm_calculator.set_coefficients(eigenpair.get_eigenvector());
+            rdms.push_back(static_cast<Eigen::MatrixXd>(rdm_calculator.calculate1RDMs().one_rdm));
+        }
+        return rdms;
+    }
+}
```
lelemmen: Add new line at end of file.

guacke:
I agree with the comments made by @lelemmen. As this PR is initially aimed at providing support for Python bindings (including _minimal_ drivers for Hubbard and FCI), I am opening an issue that concerns those comments that would take this PR too far.

guacke:
This is issue #340.

guacke:
This PR provides Python bindings support (both locally, as well as on the HPC clusters). The drivers are rudimentary (at best) as has been addressed in issue #340 . Subsequent PR's will be aimed at resolving this issue.

lelemmen:
@guacke I'm not comfortable merging this PR into develop. Since PR #341 has already been opened and was branched off from this one, I suggest that we keep this branch/PR inactive and work on PR #341 and merge that one when work is finished.

guacke:
In the hopes of resolving some Travis issues, I have rebased this branch. If the Travis builds succeed, I will clean this PR.

README.md
```diff
@@ -103,6 +103,8 @@ In general, please set and pass the following options to the `cmake ..` command:
     * the compiled library will be installed in `prefix/lib`
     * drivers (optional) and benchmarks (optional) will be installed in `prefix/bin`
     * CMake target files will be installed in `prefix/cmake`
+    
+    We note that setting `CMAKE_INSTALL_PREFIX=~/.local` is preferred as this is also makes sure that the installed Python modules can be found automatically.
```
lelemmen:
```suggestion
    We note that setting `CMAKE_INSTALL_PREFIX=~/.local` is preferred as this also makes sure that the installed Python modules can be found automatically.
```

lelemmen: Is `~/.local` automatically found by Python?

lelemmen: It seems it does: https://unix.stackexchange.com/a/407029.

Issue #329, opened 2019-06-26
Add flexibility for the direction of the H-chains


lelemmen:
Right now, we force a H-chain (or H2-chain) to be placed in the `x` direction. We could use `CartesianDirection` to let the user decide in which direction the chain should be placed. Can we switch to default `z` direction?

Issue #186, opened 2018-12-14
Implement a function that checks the validity of a given file name


lelemmen:
Bringing over this issue from cpputil

Something generalizing the following

```
// Find the extension of the given path (https://stackoverflow.com/a/51992)
    std::string extension;
    std::string::size_type idx = xyz_filename.rfind('.');

    if (idx != std::string::npos) {
        extension = xyz_filename.substr(idx+1);
    } else {
        throw std::runtime_error("I did not find an extension in your given path.");
    }

    if (!(extension == "xyz")) {
        throw std::runtime_error("You did not provide a .xyz file name");
    }

    // If the xyz_filename isn't properly converted into an input file stream, we assume the user supplied a wrong file
    std::ifstream input_file_stream (xyz_filename);
    if (!input_file_stream.good()) {
        throw std::runtime_error("The provided .xyz file name is illegible. Maybe you specified a wrong path?");
    }
```

Issue #198, opened 2018-12-14
Non-sensical compare methods for atoms


tmhuysen:
https://github.com/GQCG/gqcp/blob/b51628a3e9e758c8808af8525e835fafccd50290/src/Atom.cpp#L115

PR #339, opened 2019-07-01
Implement NuclearFramework

lelemmen:
As alluded to in #336, I need to do a small refactor: everything nuclei-based should be reflected inside a `NuclearFramework` class that has nothing to do with electrons.


lelemmen:
While implementing `NuclearFramework`, I've resolved the following issues=

1. There was no more need to check the charge of a nuclei to implement a compare functor to be able to `std::sort` `Nuclei`, so #198 can be closed.

2. I've implemented `validateAndOpen`, a function that checks if a given filename has the expected extension. This closes #186.

3. By using a given `CartesianDirection`, we can now specify the axis along which the H(2)-chains are placed. This closes #329.

PR #350, opened 2019-07-03
This closes #318.

### LINKED ISSUE #318: Readme should include --recursive
We can see that the current `git clone` command doesn't include `--recursive`. We should fix this in order to let our users know that there are submodules to be initialized.

guacke:


guacke:
Cluster installations tend to 'generate' time...

Issue #302, opened 2019-06-17
Vector as member instead of deriving from vector.


guacke:
As caught by the Intel compilers, the std::vector was not meant to be derived from: 
https://stackoverflow.com/questions/4513696/inheriting-vectors-and-initialization

As such, for instance, ShellSet should have a vector as a member instead of deriving from a member.

lelemmen:
I agree. This also closes #263.

Could you put the Intel compiler warning in this issue for later reference?

lelemmen:
This was fixed in #353.

PR #353, opened 2019-07-04
Refactor integral engines

lelemmen:
In order to treat the calculation of integrals over basis functions/shells in a uniform way, I will implement
- [x] integral engines, whose job is to calculate integrals over shells. One such calculation (over 2 or 4 shells) returns an integral buffer (see next point)
- [x] integral buffers, whose job is to place the calculated integrals (of which there are nbf_sh1 x nbf_sh2) inside the full matrix (which has dimension nbf x nbf)

The main thing that I think is necessary is compile-time checking if a certain integral engine (say libint overlap) can be combined with a shell type (say GTO shells). This is to make sure that these logical errors are caught at compile-time: libint cannot calculate integrals over gauge-including GTO shells.


lelemmen:
I forgot to `git commit --amend`, so that's why I force-pushed.

lelemmen:
8a32466 closes #302.

lelemmen:
What remains to be done is
- [ ] implement Libcint engines and buffers
- [ ] clean up and structure files within `gqcp/include/Basis`

lelemmen:
Note to the @GQCG/cpp-developers: this PR is a huge one, as was expected. Sorry! 👀

lelemmen:
It seems that Travis is tempering an error concerning the Mulliken-constrained pybindings. @tmhuysen I think it’s OK to merge this branch and then check if the resulting develop really fails. Any other thoughts?

include/Basis/Integrals/IntegralEngine.hpp
```diff
@@ -0,0 +1,148 @@
...
+// 
+// GQCG-gqcp is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with GQCG-gqcp.  If not, see <http://www.gnu.org/licenses/>.
+// 
+#ifndef GQCP_ONEELECTRONINTEGRALENGINE_HPP
```
tmhuysen: update header

tests/Basis/OneElectronIntegralEngine_test.cpp
```diff
@@ -0,0 +1,35 @@
...
+BOOST_AUTO_TEST_CASE ( sandbox ) {
+
+    const auto molecule = GQCP::Molecule::ReadXYZ("data/h2.xyz");
+
+
+    auto overlap_engine = GQCP::IntegralEngine::Libint(GQCP::Operator::Overlap());
+    auto kinetic_engine = GQCP::IntegralEngine::Libint(GQCP::Operator::Kinetic());
+    auto nuclear_engine = GQCP::IntegralEngine::Libint(GQCP::Operator::NuclearAttraction(molecule));
+
+}
```
tmhuysen: line at end of file

tmhuysen:
I've looked at the changes and do not understand where the segmentation fault can be coming from.

lelemmen:
I'm using my administrator powers to merge this PR, if Travis still fails, I will subsequently resolve the issue in a separate PR.

PR #354, opened 2019-07-04
Investigate if we can turn MKL into a 'hard' dependency.

guacke:
As LibCint needs a BLAS library, I have added the option to link to MKL (-DUSE_MKL=TRUE) on the cmake_refactor branch of the GQCP LibCint repository. However, when using this option, GQCP must _always_ be linked to MKL. Otherwise, all invocations of LibCint will fail.

One way to use GQCP without MKL, is to link LibCint to a different BLAS library than MKL. Otherwise Eigen gives strange errors when linking to MKL but not explicitly telling Eigen that you have done so. However, it seems counterintuitive to support multiple BLAS libraries as (only?) MKL is explicitly supported by Eigen. 

As such, this PR proposes to make MKL a 'hard' dependency, meaning that the option EIGEN_USE_MKL is always ON.


guacke:
This PR is superseded by PR 355.

PR #369, opened 2019-08-07
Remove dependency on Boost < 1.70.

guacke:
This PR is a follow-up for PR #368 and is intended to be able to use Boost versions higher than 1.69 (in order not to have a hard dependency on a specific version of Boost).


guacke:
Apparently, this is a CMake/Boost issue: https://github.com/ecmwf/ecbuild/issues/11. I will investigate further.

PR #372, opened 2019-08-07
Add additional layer in headers (resolves #292).

### LINKED ISSUE #292: Additional hierarchy layer in headers
Should we move the headers inside a 'gqcp' directory? If we do this, then our include statements become:

```
#include "gqcp/Basis/ShellSet.hpp"
```
This would reduce the pollution of the install include directory (i.e. /usr/local/include/version.hpp vs /usr/local/include/gqcp/version.hpp) and is more in line with imports from e.g. Eigen and Boost.

guacke:
This PR makes sure that headers are installed in `${INSTALL_PREFIX_PATH}/include/gqcp`. Hence, in accordance with the Eigen and Boost conventions, includes should have the form `#include "gqcp/gqcp.hpp".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,6 +184,7 @@ install(
         ${CMAKE_BINARY_DIR}/cmake/gqcp-config-version.cmake
         ${CMAKE_SOURCE_DIR}/cmake/FindEigen3.cmake
         ${CMAKE_SOURCE_DIR}/cmake/FindInt2.cmake
+        ${CMAKE_SOURCE_DIR}/cmake/FindMKL.cmake
         DESTINATION cmake
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,11 +149,11 @@ install(TARGETS gqcp
         LIBRARY DESTINATION lib
         ARCHIVE DESTINATION lib
         RUNTIME DESTINATION bin
-        INCLUDES DESTINATION include
+        INCLUDES DESTINATION include/gqcp
         )
 
-install(DIRECTORY ${CMAKE_SOURCE_DIR}/include/ DESTINATION include)
-install(FILES ${CMAKE_BINARY_DIR}/include/version.hpp DESTINATION include)
+install(DIRECTORY ${CMAKE_SOURCE_DIR}/include/ DESTINATION include/gqcp)
+install(FILES ${CMAKE_BINARY_DIR}/include/version.hpp DESTINATION include/gqcp)
 
 install(EXPORT gqcp-targets
         FILE gqcp-targets.cmake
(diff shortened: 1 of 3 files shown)


tmhuysen:
I was under the impression that the gqcp.hpp file would still be in the include folder only the remainder of the includes would be in the include/gqcp

lelemmen:
> in accordance with the Eigen and Boost conventions, includes should have the form `#include "gqcp/gqcp.hpp".

@guacke This is only for external libraries and executables, right? This shouldn't change our internal includes, right?


> I was under the impression that the gqcp.hpp file would still be in the include folder only the remainder of the includes would be in the include/gqcp

@tmhuysen I guess it would indeed be nice to have a hierarchy like this:
* `${INSTALL_PREFIX_PATH}$
    * `gqcp.hpp`
    * `gqcp`
        * all the gqcp includes

guacke:
@tmhuysen @lelemmen If we want to impose the hierarchy
* `${INSTALL_PREFIX_PATH}$
    * `gqcp.hpp`
    * `gqcp`
        * all the gqcp includes

I guess that we
- [ ] need to adjust `gqcp.hpp` such that its imports are relative to gqcp. E.g. `#include "gqcp/Basis/AOBasis.hpp"` instead of `"#include "Basis/AOBasis.hpp"`.
- [ ] Impose this hierarchy (i.e. directory structure) also for internal includes, which should involve some heavy refactoring.

The current refactor only adds an additional layer to the gqcp imports for external libraries.

lelemmen:
I would say that the changes that @guacke summarized would be for the best.

guacke:
Can I delegate those refactoring steps to @tmhuysen?

lelemmen:
@guacke you can assign the issue and PR to me, as I’ll be coding next week.

lelemmen:
@guacke In conjunction with the current [gqcp-link](https://github.com/GQCG/gqcp-link/tree/5f0e8187a2d4111afde781135638fe49deb60429), it doesn't seem necessary to refactor the internal names.

lelemmen:
@GQCG/cmake-developers Could you review this PR? Since the inclusion of gqcp-link was incorporated into develop after this branch was branched off, I'll fix the gqcp-link issues later.

guacke:
@lelemmen So we no longer require gqcp.hpp to live outside include/gqcp? @GQCG/cmake-developers I cannot review this PR, as I opened the PR.

lelemmen:
> @lelemmen So we no longer require gqcp.hpp to live outside include/gqcp? @GQCG/cmake-developers I cannot review this PR, as I opened the PR.

Indeed, `gqcp.hpp` (the collective include header) is installed in `prefix/include/gqcp/`, so external projects should now include it through `<gqcp/gqcp.hpp>`.

lelemmen:
The failing test is fixed in #401.

PR #374, opened 2019-08-07
Refactor Boost support.

guacke:
This PR is a follow-up for PR #368 and is intended to be able to use Boost versions higher than 1.69 (in order not to have a hard dependency on a specific version of Boost).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,8 +86,7 @@ target_link_libraries(gqcp
             Spectra::Spectra
             MKL::MKL
         )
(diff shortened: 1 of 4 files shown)


guacke:
As the cmake support of the 1.70 release is currently bugged (see e.g. https://github.com/ecmwf/ecbuild/issues/11), we currently only support Boost <= 1.69.

guacke:
This PR ended up refactoring the CMake invocations, as the current difficulties with Boost 1.70 seem to be situated upstream. The documentation now reflects this current dependency.

PR #401, opened 2019-08-21
Fix Travis

### LINKED ISSUE #399: Speed up the calculation of integrals - Libcint
We have `LibcintIntegralEngine`s for one- and two-electron operators that implement the calculation of integrals over a pair or quartet of shells.

As stated in the [documentation of Libcint](https://github.com/sunqm/libcint), it is possible to:
- [x] skip shell calculations if they are `not0` (implemented in #401);
- [ ] ~add optimizer structs for the calculation of the two-electron integrals~ (see #401).

**Additional context**
This is fundamentally different from #67, because it has nothing to do with the looping over the shells.

lelemmen:
**Short description**
Travis currently fails: this PR aims to resolve this.

**Longer description**
The error that Travis reported lead us to a bug that was introduced in #353. As it was related to some integrals being zero in the Libint2 calculations, I implemented a 'screening' function for the Libint buffers, which is a function that checks if all the integrals calculated over two or four shells are zero. Since the extension for the Libcint buffers and the implementation of the two-electron optimizers were straightforward (and my memory of Libcint was still fresh), this PR also closes #399.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,5 +99,5 @@ script:
   - git clone https://github.com/GQCG/gqcp-link.git
   - cd gqcp-link
   - mkdir build && cd build
-  - cmake .. -DCMAKE_PREFIX_PATH=~/.local/cmake -DCMAKE_MODULE_PATH=~/.local/cmake
+  - cmake .. -DCMAKE_PREFIX_PATH=${HOME}/miniconda -DGQCP_INSTALL_PREFIX=~/.local
   - make -j3 && ./test_driver
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,11 +144,11 @@ install(TARGETS gqcp
         LIBRARY DESTINATION lib
         ARCHIVE DESTINATION lib
         RUNTIME DESTINATION bin
-        INCLUDES DESTINATION include
+        INCLUDES DESTINATION include/gqcp
         )
 
-install(DIRECTORY ${CMAKE_SOURCE_DIR}/include/ DESTINATION include)
-install(FILES ${CMAKE_BINARY_DIR}/include/version.hpp DESTINATION include)
+install(DIRECTORY ${CMAKE_SOURCE_DIR}/include/ DESTINATION include/gqcp)
+install(FILES ${CMAKE_BINARY_DIR}/include/version.hpp DESTINATION include/gqcp)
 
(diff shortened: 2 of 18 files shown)


lelemmen:
I think I've found what causes the error. In #353, I removed the behavior in the following lines:

```cpp
 if (calculated_integrals == nullptr) {  // if the zeroth element is nullptr, then the whole shell has been exhausted
    // or the libint engine predicts that the integrals are below a certain threshold
    // in this case the value does not need to be filled in, and we are safe because we have properly initialized to zero
    continue;
}
```

I will implement a solution to this bug.

lelemmen:
71f722 implements part of #399.

lelemmen:
The current failures of Travis are due to #372 not being merged yet.

lelemmen:
The failing test is fixed in #401.

lelemmen:
@GQCG/cpp-developers Could you review this branch? It's bound to fail because of Travis (see my previous comment), the resulting develop branch should be OK.

guacke:
@lelemmen #372 has been merged. Could you rebase this branch on the current develop to ensure Travis compatibility?

lelemmen:
cca65c2 implements Libcint two-electron optimizers, so it closes #399.

lelemmen:
Even the most trivial change, e.g. inside `LibcintTwoElectronEngine.calculate()`:

```cpp
CINTOpt* opt;
cint2e_cart_optimizer(&opt, this->libcint_raw_container.atmData(), this->libcint_raw_container.numberOfAtoms(), this->libcint_raw_container.basData(), this->libcint_raw_container.numberOfBasisFunctions(), this->libcint_raw_container.envData());

const auto result = this->libcint_function(libcint_buffer, shell_indices, this->libcint_raw_container.atmData(), this->libcint_raw_container.numberOfAtoms(), this->libcint_raw_container.basData(), this->libcint_raw_container.numberOfBasisFunctions(), this->libcint_raw_container.envData(), opt);

CINTdel_optimizer(&opt);
```

I'm reporting the memory access violations. I would therefore conclude that there is something wrong with (the linking with?) Libcint.

I'm reverting the implementation of the optimizer structs.

lelemmen:
Since nothing extra was implemented after @tmhuysen's review, I will accept his approvement and merge this PR.

PR #405, opened 2019-08-22
Implement block representations

### LINKED ISSUE #323: Implement block matrix/tensor representations
In #313, I've implemented two `Block` classes, `BlockMatrix` and `BlockRankFourTensor`, that behave like the mathematical quantities `G_i^a` (geminal coefficients like in the notes) and `J_{ia, jb}` (elements of the Jacobian of the PSEs). However, I'm not sure about the naming: I think we can decide on this when we're doing the next representation refactor (as we have alluded to in #314).

lelemmen:
**Short description**
In this PR, I will implement some necessary block matrix representations, as alluded to in #323.

**Related issues**
This PR closes #323.

**Don't forget:**


lelemmen:
> Nothing wrong with the PR. Just wondering if there is something you don't like with what eigen3 provides?

@tmhuysen There's no problem with Eigen, it's just that I want the `operator()` to behave like it does in the notes, and that is not what Eigen provides. Thanks for the review!

PR #406, opened 2019-08-23
Refactor second-quantized operators

### LINKED ISSUE #309: Generalize Operator to work with multiple components
Right now, the electric dipole operator is represented as a `std::array` of `OneElectronOperator`s. We should generalize this, to allow the creation of operators with multiple (Cartesian) components: we should fully implement the concept of 'vector operators'. Their 'scalar' term should then also be a vector.

### LINKED ISSUE #310: Move Fockian calculations to Operators
Right now, only `HamiltonianParameters` knows how its Fockian (and super-Fockian) matrix is calculated. Since we have derived the theory on how to calculate Fockians for one- and two-electron operators, it will be better if they both know how their Fockians are calculated.

### LINKED ISSUE #384: Second-quantized operators
This issue will gather all thoughts I've had on the introduction of the so-called `SQOperator`, or 'second-quantized' operator: 
- [x] `OneElectronOperator` and `TwoElectronOperator` should be immediately renamed to `SQOneElectronOperator` and `SQTwoElectronOperator` because they are the second-quantized forms of one- and two-electron operators. (done in #406.)

- [x] `SQOperator` could benefit from a templatization as `SQOperator<Scalar, Components>`, where `Scalar` is the scalar representation of one of its matrix elements because its underlying matrix elements also are templated with `Scalar` and `Components` is the number of components the operator has (e.g. 1 for a Coulomb-type and 3 for a dipole-type).

lelemmen:
**Short description**
In this PR, I will implement some maintenance modifications for second-quantized operators.

**Related issues**
This PR:
- [x] closes #309;
- [x] closes #310;
- [x] closes #384.

**Don't forget:** 


lelemmen:
@guacke I realize that @tmhuysen is unavailable, so could you review this PR? The main thing I implemented is covered in the issues that are mentioned in the PR description.

PR #419, opened 2019-08-30
Single-particle basis

### LINKED ISSUE #333: Single particle basis
As discussed during the code review, the code could benefit from a new class: `SPBasis` short for `SingleParticleBasis`.

I would like to start by introducing the concept of a `ScalarBasis`: it is a scalar orbital basis in which integrals can be calculated. 
- [x] I would replace the current `AOBasis` by `ScalarBasis<GTOShell>` , but in principle we could also implement `GIAOBasis`, `PlanarBasis` or other types.
- [x] `ScalarBasis` should implement `operator(size_t)`, which returns a `LinearCombination` of its underlying `BasisFunctions`. Therefore, I think `ScalarBasis` should be templated as `ScalarBasis<BasisFunction>`, where `BasisFunction` can be `CartesianGTO` or `GIAO` or something else. (As alluded to in #360, we will use Cartesian GTOs to implement `operator()(r)`, but not spherical GTOs. Spherical GTOs can however be used in the calculation of integrals.)

An `SPBasis` does not need to know what kind of scalar basis underlies it, it should work with any underlying scalar basis. However, in order to implement `SQOperator` (see #384) that is able to check at compile-time if the operator integrals can be calculated, `SPBasis` would benefit from the templatization `SPBasis<ScalarBasis>`. `SPBasis` should be able to:
- [x] get the total transformation matrix since the construction
- [x] transform, rotate, ... given a transformation matrix
- [x] return the matrix representation of a given operator, using the interface discussed in #411.

### LINKED ISSUE #411: Quantization of first-quantized operators
Right now, we ask an `AOBasis` to calculate the integrals.

After having introduced an `SPBasis` (see issue #333), in order to create such a second-quantized operator, I would like to introduce the API

```cpp
Operator.quantize(SPBasis<ScalarBasis>)
```

which returns the second-quantized representation of the operator in the given single-particle basis. We should note that the returned object **is** a `SQOperator`, which should **hold** its matrix representation (i.e. its integrals). `Scalar` can be determined by `SPBasis::Scalar * ScalarBasis::Scalar`.

**Additional context**
This should be only addressed after #333, in order to keep the work done in one PR as small as possible.

lelemmen:
**Short description**
This PR will implement the beginnings of `SPBasis` and `ScalarBasis`, which serve to replace the current `AOBasis` and are necessary to start #411.

**Related issues**
This PR:
- [x] closes #333;
- [x] closes #411;

**Don't forget:** 


lelemmen:
@GQCG/cpp-developers As stated in the PR description, `ScalarBasis<GTOShell>` is the new `AOBasis`, and `SPBasis` will in a subsequent PR be used for the `FQOperator.quantize(SPBasis)` interface.

lelemmen:
This PR closes #411 because I have introduced the `quantize()` API. However, I'm not using the API as in #411, but rather

```cpp
SPBasis.quantize(FQOperator)
```

because I feel it's the single-particle basis that actually **does** the second-quantizing.

PR #427, opened 2019-09-06
Hamiltonian refactor

### LINKED ISSUE #385: Hamiltonian refactor
As alluded to in #307, `HamiltonianParameters` is a pretty unusual class, it's more like a collection of second-quantized operators, which can each be one-electron or two-electron operators. 

Maybe this is the way to go: `SQHamiltonian` is merely a collection of `SQOperators` (see #384), that adheres to the `SQOperator` methods? `SQHamiltonian` would be the current `HamiltonianParameters`. It could then also have the methods `.core()` which returns the total one-electron part and `.twoElectron()` that returns the total two-electron part? I'm thinking of just keeping a list of `SQOneElectronOperator`s and `SQTwoElectronOperator`s for all the specific interactions, although it might not be always necessary to keep the different kinds of contributions.

We could do something like `FQHamiltonian.add(ScalarOperator)` to specify the scalar operators that are used in the Hamiltonian. Using `ScalarOperator` has the requirement that the matrix-matrix product for example an electric field and the dipole operator should return such a `ScalarOperator`.

There would also be no more need keep `BaseHamiltonianParameters`.

This issue is of course related to #403.

### LINKED ISSUE #403: Consistency of transformation matrices
In #307, it was alluded that

> Every operator should know this because every operator (representation) has an underlying orbital basis.

After some thinking, this revealed that the deeper, underlying issue is to maintain consistency between the matrix representations of `SQOperator`s and `SPBasis`: transforming a basis means that the second-quantized operators expressed in it should also change. 

What has been bugging me is the example of model Hamiltonians, like Hubbard, that we do not (explicitly) express in a single-particle basis: we only let them contain the second-quantized 'integrals'.

If the `SQOperator`, `SPBasis` and `Hamiltonian` refactor is complete, we can maintain this consistency by implementing a user-friendly function:

```cpp
void transform(1e, 2e);
void transform(1e, 2e, spbasis);
```

that can be used to transform one-electron operators, two-electron operators and single-particle bases simultaneously. We can even implement `void transform(ham, spbasis)` to make it easier for the user/developer, but you get the gist of it.

Ultimately, this means that we enforce that `SPBasis` is the interface class that you can ask the total transformation matrix of, and that we don't give `SQOperator` the interface to ask its underlying basis: that is the job of the user/developer to know.

**Additional context**
This issue replaces and supersedes #307, because it was poorly worded and not sufficiently clear. 
This issue is related to the operator refactor in #385, #333 and #384.

lelemmen:
**Short description**
This PR will implement the necessary changes to support the Hamiltonian refactor.

**Related issues**
This PR:
- [x] closes #385;
- [x] closes #403.

**Additional context**
This will be a huge refactor, which is why I will approach this as follows. First I will add the necessary new functionality, and only afterwards I will delete the old interfaces. This is to make sure that the code stays compilable during the whole refactor.

**Don't forget:** 


lelemmen:
@tmhuysen So I managed to track down this pesky little bug 🐞! In the new code, I had defined the Löwdin-orthogonalization matrix as `T=S_AO^{-1/2}` instead of taking into account the current transformations that had taken place. However, I think the correct Löwdin orthonormalization matrix is `T=S_current^{-1/2}`, which is exactly what the code used to do.

That said, it's peculiar that using the `S_AO^{-1/2}` basis gives me an FCI energy that is lower than your reported calculations. However, it's not that surprising, since you are doing frozen core calculations, right? 
Maybe this behavior of the FCI Mulliken code is unintended: if the code fails a regular DIIS, but manages the DIIS on the fragments, it will **always** transform to your special `T`-matrix (that combines the fragments). If the DIIS using `T` as initial orbitals fails, your code proceeds by Löwdin orthonormalizing, but since the orbitals had already been transformed with `T`, the Löwdin orthonormalization does not transform the orbitals to the `S_AO^{-1/2}` basis, but to `(S_AO*T)^{-1/2}`, in which apparently the frozen core FCI calculation has a higher energy.

PR #437, opened 2019-09-26
Geminals refactor

### LINKED ISSUE #324: Refactor geminals
Bringing over from #313:

- [x] implement `AP1roGPSEs` and make sure that the correct (now: column-major instead of the previous row-major) representations are used for the vectors and matrices. This is mainly just bringing over the old code.
- [x] use `Block` representations for `APIGVariables`

lelemmen:
**Short description**
In this PR, I will refactor some of the geminals classes such that their use is optimized.

**Related issues**
This PR:
- [x] closes #324

**Don't forget:** 


lelemmen:
I have now refactored both AP1roG-related solvers: `AP1roGPSESolver` (which solves the PSEs) and `AP1roGLagrangianOptimizer` (which solves for the Lagrangian multipliers). 

What should be done next is the following:
- [ ] Bring over the old code with corrected formulas for the AP1roG PSE Jacobian and better use of the `Block` representations
- [ ] As described in #324, use `Block` representations for APIG-related variables
- [ ] Move `FromRowMajor` and `FromColumnMajor`, currently in `AP1roGGeminalCoefficients` to `AP1roGVariables`

src/Geminals/AP1roGLagrangianOptimizer.cpp
```diff
@@ -17,51 +17,72 @@
...
+    VectorX<double> x = linear_solver.solve(-b);
 
-    if (std::abs((J * x - b).norm()) > 1.0e-12) {
+    if (std::abs((J * x + b).norm()) > 1.0e-12) {
         throw std::runtime_error("void AP1roGLagrangianOptimizer::solve(): The Householder QR decomposition failed.");
     }
 
-    this->multipliers = AP1roGVariables(x, this->N_P, K);
+
+    return AP1roGVariables(x, N_P, K);  // specify: column major or row major
```
lelemmen: - [ ] specify column major or row major

lelemmen:
As the reviewers might notice, I've also changed the API for `NewtonNLSysEqSolver` in a way that I think it natural: 
- a modifiable argument that, on input is an initial guess and on output is the converged solution
- if no initial guess is provided, a zero initial guess is used and the result is returned

src/Geminals/AP1roGGeminalCoefficients.cpp
```diff
@@ -69,32 +97,76 @@ AP1roGGeminalCoefficients::AP1roGGeminalCoefficients(size_t N_P, size_t K) :
...
-    for (size_t mu = 0; mu < number_of_geminal_coefficients; mu++) {
-        size_t i = GQCP::matrixIndexMajor(mu, K, N_P);
-        size_t a = GQCP::matrixIndexMinor(mu, K, N_P);
-
-        g_vector(mu) = - g(a,i,a,i) / (2 * (h(a,a) - h(i,i)));
+    BlockMatrix<double> G (0, N_P, N_P, K);
+    for (size_t i = 0; i < N_P; i++) {
+        for (size_t a = N_P; a < K; a++) {
+            G(i,a) = - g(a,i,a,i) / (2 * (h(a,a) - h(i,i)));
+        }
```
lelemmen: For the reviewers: I think this is an excellent example of the cleaner code when using `BlockMatrix`.

PR #445, opened 2019-10-28
Feature/constrained spin

tmhuysen:
**Short description**

- [x] Adds support for an unrestricted framework. This PR will allow for FCI to have different integrals for beta and alpha components.
- [x] Ability to calculate S<sub>z</sub> ~and S<sup>2</sup>~.
- [x] Update constrained pybinding

To accommodate the existing pybinding for MullikenConstrainedFCI, unrestricted frozencore FCI was implemented and to test this properly unrestricted selected CI was also implemented. 

**Related issues**
GQCG/constrained_outline#4

**Additional context**
This serves potentially as beginnings for an unrestricted framework and does not include expansion of all implemented methods, only what is required correctly implement the spin constraints.

**Don't forget:** 


lelemmen:
I think this is indeed a good way to progress. In a first step, implement what is needed for research, like in your case classes that can be used in unrestricted calculations. The second step then consists of refactoring into a more elegant solution if it ever comes up to mind at a later stage.

lelemmen:
- [ ] As discussed, could you also implement an `USpinorBasis`, with suitable functionality/interface for unrestricted spinor bases?

tmhuysen:
I will do USpinorBasis after the GSpinorBasis PR is finished, rather than try to work it in here, everything I need is in place, and is technically correct code, I feel like it would be chaotic to both do a PR on SpinorBasis simultaneously.

lelemmen:
Again, a lot has changed only in one PR: 
- `USQHamiltonian`
- Unrestricted flavors of CI
- Spin-related Mulliken partitioning
- Your Mulliken bindings

Please, for my reviewing sanity, break big PRs down into smaller ones.

tmhuysen:
This PR is being spit into to multiple smaller PRs

PR #473, opened 2020-01-14
Restructure GQCP

lelemmen:
**Short description**
In this PR, I've restructured our code package. More information can be found on the [GQCP web page](https://gqcg.github.io/GQCP/).

**Don't forget:** 


lelemmen:
@GQCG/cpp-developers Here's a small message to the reviewers. Even though this is a +1000/-1000 PR, the only real thing that happened in this PR is changing the folder structure of the GQCP.

PR #480, opened 2020-01-24
Dipole operator Python bindings

### LINKED ISSUE #414: Move expectation values to SQOperator?
Currently, expectation values are calculated in a separate function.

I think it would make more sense to add the calculation of expectation values to the `SQOperator`s.

**Additional context**
It would be best to implement this only after #385, to avoid duplicate work.

lelemmen:
**Short description**
In this small PR, I will implement the necessary Python bindings such that expectation values over the dipole operator can be calculated.

**Related issues**
This PR will:
- fix #414

**Don't forget:** 


lelemmen:
Closing this PR in favor of #508.

PR #514, opened 2020-02-13
GSpinorBasis extensions

### LINKED ISSUE #215: Calculate spin expectation values
An implementation of the expectation values `S_x`,  `S_y`, `S_z` and `S^2` is connected with the underlying spinor basis: the one- and two-electron 'integrals' can be calculated (if my understanding is complete) using the spinor basis. The expectation value can then be calculated as a contraction with the corresponding 1- or 2-DM.

I would say that every spin operator could become a `FQOperator`, which can be `quantized()`, using a `SingleParticleBasis` into an `SQOperator`.

### LINKED ISSUE #449: Calculate integrals over different shell sets
In order to calculate the (alpha-beta)-block of the matrix representation of a one-electron operator in the AO basis, **we have to be able to calculate integrals over different `ShellSet`s** if we want to maintain full generality with respect to expanding the alpha- and beta-components in different scalar bases. 

Right now, the API to calculate integrals is 

https://github.com/GQCG/GQCP/blob/efe2058979bd053a372f2e80c929cf5a96be00a1/include/Basis/Integrals/IntegralCalculator.hpp#L52-L53

where we see that we can only calculate integrals where the left and right shell come from the same `ShellSet`. This should be generalized to accept different `ShellSet`s, which shouldn't be difficult.

Furthermore, the higher-level API for calculating integrals is then for example

https://github.com/GQCG/GQCP/blob/efe2058979bd053a372f2e80c929cf5a96be00a1/include/Basis/ScalarBasis.hpp#L144

which is sufficient for the alpha-alpha and beta-beta blocks, but should be changed if we want to calculate alpha-beta and beta-alpha blocks. Since we will have to effectively combine two scalar bases, I think the API will have to change.

**Additional context**
If we have implemented this, we can calculate matrix elements over the spin-orbit coupling operator (if we want to add some scalar relativistic effects) and we can then also do calculations in constant magnetic fields with an arbitrary direction.

### LINKED ISSUE #512: Two electron integrals in generalised spinor basis
@lelemmen 

A possibility to get the two electron integral tensor in generalised spinor basis.

**Additional context**
This would be a very useful addition, in order to be able to perform Hessian calculations in a generalised basis.

lelemmen:
**Short description**
In this PR, `GSpinorBasis` will receive a couple of extensions/new features:
- [x] Python bindings for `GSpinorBasis`

**Related issues**
This PR will:
- [x] implement and close #449, but only for Libint2 engines;
- [x] implement the quantization of two-electron operators, so it closes #512.
- [x] implement the quantization of the spin operator `GSpinorBasis`, so it closes #215.

**Don't forget:** 


lelemmen:
@GQCG/cpp-developers I'm using my admin powers to merge this one.

PR #546, opened 2020-02-25
Add EasyConfig.

guacke:
**Short description**
Enable EasyBuild support for GQCP.

**Additional context**
When we have introduced a release system, an updated EasyConfig should be pushed to the official EasyConfig repo and a system wide install by the HPC support staff should be requested.


lelemmen:
@guacke Could you give us an update on this branch? Is it finished? Should it be closed?

guacke:
@lelemmen I am restarting work on this PR.

guacke:
This PR is failing on the cluster due to a mismatch with the Boost Preprocessor used in Libint.

```
In file included from /data/gent/vo/000/gvo00004/easybuild/CO7/skylake-eth/software/Boost/1.69.0-intel-2019b/include/boost/optional/optional.hpp(56),
                 from /data/gent/vo/000/gvo00004/easybuild/CO7/skylake-eth/software/Boost/1.69.0-intel-2019b/include/boost/optional.hpp(15),
                 from /data/gent/vo/000/gvo00004/easybuild/CO7/skylake-eth/software/Boost/1.69.0-intel-2019b/include/boost/format/internals.hpp(21),
                 from /data/gent/vo/000/gvo00004/easybuild/CO7/skylake-eth/software/Boost/1.69.0-intel-2019b/include/boost/format.hpp(38),
                 from /tmp/vsc40558/gqcp/dev/intel-2019b-Python-3.7.4/gqcp/gqcp/src/Basis/ScalarBasis/CartesianExponents.cpp(20):
/data/gent/vo/000/gvo00004/easybuild/CO7/skylake-eth/software/Boost/1.69.0-intel-2019b/include/boost/utility/result_of.hpp(221): catastrophic error: cannot open source file "libint2/boost/preprocessor/iteration/detail/iter/forward1.hpp"
  #include BOOST_PP_ITERATE()
```

Will investigate further.

guacke:
@lelemmen This is related to issue #547.

guacke:
The libint library overwrites Boost macros when used; this leads to internal errors in the Boost libraries which depend on those macros. A very hacky fix that _seems_ to work on the UGent HPC is deleting the libint2/boost include dir after installation, but we have no indication what is happening internally and how Libint is corrupting the internal settings of Boost. 

If we **really** want to keep using libint, this essentially forces us to consider discontinuing all use of the Boost libraries in GQCP. As we don't really need the program options, only the testing framework remains. A refactor to Google Test would be a viable option.

Another option would be to switch entirely to Libcint, but this library also has shown to be problematic (see current density research). However, this would enables us to keep using Boost. 

@lelemmen What are your thoughts?

guacke:
This PR adds support for building GQCP on the UGent HPC infrastructure using EasyBuild. Note that the Libint 2.4.2 dependency refers to a _hacked_ version where the Boost preprocessor dir has been removed from the include dirs of Libint.

lelemmen:
> @lelemmen This is related to issue #547.

Do you think the issue can be resolved if we migrate to Libint 2.6.0+?

> The libint library overwrites Boost macros when used; this leads to internal errors in the Boost libraries which depend on those macros. A very hacky fix that _seems_ to work on the UGent HPC is deleting the libint2/boost include dir after installation, but we have no indication what is happening internally and how Libint is corrupting the internal settings of Boost.

Yeah, that's just ... I'm not sure what the libint developers were thinking in this case ... I'm glad that this workaround might be suitable, though.

> If we **really** want to keep using libint, this essentially forces us to consider discontinuing all use of the Boost libraries in GQCP. As we don't really need the program options, only the testing framework remains. A refactor to Google Test would be a viable option.

While I agree that using a different testing framework could indeed work, I would like to add that we're using boost in other ways than only its testing framework. We're using mathematical functions, constants, formatting, etc. just to name a few.

> Another option would be to switch entirely to Libcint, but this library also has shown to be problematic (see current density research). However, this would enables us to keep using Boost.

You know my opinion about libcint. I couldn't even find the correct one-electron function in its own documentation, but luckily pyscf has some documentation that can be helpful.


As a conclusion: I'm not happy with either Libint or Libcint, but until we have our own suite of integral engines, I'm not sure we can remove either dependency.

guacke:
This PR addresses #597 and part of #596.

Issue #558, opened 2020-03-12
Unrestricted types


lelemmen:
**Describe the feature you'd like**
As a first milestone in order to implement UHF, we should tackle the problem of how to encapsulate objects that are of 'unrestricted' type. 

- [x] This would mean that we should make sure that `USpinorBasis` is linked to `USQOneElectronOperator`,  `USQTwoElectronOperator` and `USQHamiltonian`, following an API that is similar to what `RSpinorBasis` provides.

lelemmen:
@xdvriend I think that it's a good start for you to work on this issue before implementing UHF SCF. This will make sure that you learn to work with the internals of the code first, and afterwards UHF SCF will be quite parallel to RHF SCF.

xdvriend:
I will take this step by step. First, i will try and fix the USQ Operators one by one.

- [ ] USQOneElectronOperator #561 

- [ ] USQTwoElectronOperator

- [ ] USQHamiltonian

PR #566, opened 2020-04-01
Feature/usq hamiltonian

xdvriend:
**Short description**
The goal of this PR is to rewrite the USQ Hamiltonian in a way that it uses the newly implemented USQ operators, in a way similar to how the SQ Hamiltonian uses the SQ operators.

**Related issues**
#558


lelemmen:
@xdvriend I'll close this PR since it's inactive at the moment, but I'll keep the branch available on this repository so that we can start working from it later.

PR #572, opened 2020-04-24
Fix std::vector bugs

lelemmen:
**Short description**
In this PR, I will fix recently introduced `std::vector` bugs.

**Related issues**

**Additional context**

**Don't forget:** 

### DIFF
--- a/gqcp/src/ONVBasis/SeniorityZeroONVBasis.cpp
+++ b/gqcp/src/ONVBasis/SeniorityZeroONVBasis.cpp
@@ -75,9 +75,7 @@ VectorX<double> SeniorityZeroONVBasis::evaluateOperatorDiagonal(const ScalarSQOn
     }
 
     // Prepare some variables to be used in the algorithm.
-    const auto N_P = this->numberOfElectronPairs();
     const auto dim = this->dimension();
-
     const auto& f = one_op.parameters();
 
     VectorX<double> diagonal = VectorX<double>::Zero(dim);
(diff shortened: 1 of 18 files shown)


gqcp/src/ONVBasis/SeniorityZeroONVBasis.cpp
```diff
@@ -67,8 +67,35 @@ size_t SeniorityZeroONVBasis::calculateDimension(const size_t K, const size_t N_
...
+    this->forEach([&diagonal, &f](const SpinUnresolvedONV& onv, const size_t I) {
+        double value = 0;  // to be added to the diagonal
+
+        // Loop over occupied orbital index and add the contribution due to the spinor that it occupies.
+        onv.forEach([&value, &f](const size_t p) {
+            value += 2 * f(p, p);  // *2 because of seniority-zero
+        });
+
+        diagonal(I) += value;
+    });
```
lelemmen, line 96: @guacke @xdvriend @dariatols @nbilliet What do you think of this source code, related to the following older implementation?

https://github.com/lelemmen/GQCP/blob/08cddab57aba3fe3746cf587a9d9b86dee5d9d50/gqcp/src/QCMethod/CI/HamiltonianBuilder/DOCI.cpp#L200-L227

dariatols: There are less for-loops to keep track of in the newer code, so that makes it easier to read. The comment `Iterate over every proxy doubly-occupied ONV.` was not enough for me to explain what `forEach` exactly does, but the body of `forEach` explains itself.

lelemmen:
@GQCG/cpp-developers I'm using my admin powers to merge this PR before reviewing.

PR #574, opened 2020-05-01
Feature/ccd0

dariatols:
**Short description**
Implementation for CCD0.

**Related issues**
https://github.com/GQCG/CCD0/issues/6

**Don't forget:** 


lelemmen:
Hi @dariatols, could you please remove the CMake build files? Since there are thousands of new files,I can’t find which files are actually changed.

dariatols:
@lelemmen This is done now! Should I add these folders (`CMakeFiles/` etc) to the .gitignore file?

The main code is written in `gqcp/tests/QCMethod/CC/CCD0_test.cpp` for now, it makes easier to run the file directly.

lelemmen:
@dariatols Thanks for cleaning up your branch!

No, I think you might have once done an in-source build. If you out-of-source compile the source code in a `build` directory, they won’t be staged/committed because these folders are already in the `.gitignore`.

I’ll look into the code you’ve written later.

gqcp/include/QCMethod/CMakeLists.txt
```diff
@@ -10,3 +10,4 @@ add_subdirectory(Geminals)
 add_subdirectory(HF)
 add_subdirectory(OrbitalOptimization)
 add_subdirectory(RMP2)
+add_subdirectory(CC)
```
lelemmen, line 13: Could you list this alphabetically?

gqcp/tests/QCMethod/CC/CCD0_test.cpp
```diff
@@ -0,0 +1,257 @@
+#define BOOST_TEST_MODULE "CCD0"
+
+#include <boost/test/unit_test.hpp>
```
lelemmen, line 3: Please add the header files from ‘more specific’ to ‘more general’.

This means
- Our includes first
- Boost/Eigen headers
- STL headers


Unless it’s headers for `boost/test`, these should always go first because Libint (one of our dependencies otherwise starts complaining).

gqcp/tests/QCMethod/CC/CCD0_test.cpp
```diff
@@ -0,0 +1,257 @@
...
+#include "Operator/SecondQuantized/SQOneElectronOperator.hpp"
+#include "Operator/SecondQuantized/SQTwoElectronOperator.hpp"
+#include "QCMethod/HF/RHFSCFEnvironment.hpp"
+#include "QCMethod/HF/RHFSCFSolver.hpp"
+#include "QCMethod/HF/DiagonalRHFFockMatrixObjective.hpp"
+#include "QCMethod/HF/RHF.hpp"
+#include "QCMethod/QCStructure.hpp"
+#include "Mathematical/Representation/BlockRankFourTensor.hpp"
+
+GQCP::QCRankFourTensor<double> twoElectronAntiSymmetrized(const GQCP::QCRankFourTensor<double>& gs) {
```
lelemmen, line 28: We often name the operators `g_op` and then we use the parameters as `const auto& g_par = g_op.parameters()`.

gqcp/tests/QCMethod/CC/CCD0_test.cpp
```diff
@@ -0,0 +1,257 @@
...
+                }
+            }
+        }
+    }
+    return ls;
+}
+
+double calculateProjection(const GQCP::SQHamiltonian<double>& sq_hamiltonian, GQCP::BlockRankFourTensor<double>& ts, const GQCP::ScalarSQOneElectronOperator<double>& f,
+                        const size_t a, const size_t b, const size_t i, const size_t j){
+    const auto hs = sq_hamiltonian.core().parameters(0);
```
lelemmen, line 46: Since the parameters themselves exist already in memory, it’s better to add a ‘reference’ here: ‘const auto& h = sq_hamiltonian.core().parameters()’.

Furthermore, it seems that your amplitudes are not modified: could you add a ‘const’ to indicate this?

gqcp/tests/QCMethod/CC/CCD0_test.cpp
```diff
@@ -0,0 +1,257 @@
...
+    const auto gs = sq_hamiltonian.twoElectron().parameters(0);
+    const auto ls = twoElectronAntiSymmetrized(gs);
+    const auto fs = f.parameters();
+
+    const size_t N = gs.dimension(); // returns the number of electrons
```
lelemmen, line 51: This does not return the number of electrons, but rather the number of orbitals the operator is expressed in. In your case, this is `K`, the number of spatial orbitals. You’ll have to find another way to specify the number of electrons/electron pairs.

gqcp/tests/QCMethod/CC/CCD0_test.cpp
```diff
@@ -0,0 +1,257 @@
...
+            }
+        }
+    }
+    
+    
+    return projection;
+}
+
+GQCP::BlockRankFourTensor<double> calculateProjections(const GQCP::SQHamiltonian<double>& sq_hamiltonian, GQCP::BlockRankFourTensor<double>& ts, const GQCP::ScalarSQOneElectronOperator<double>& f){
+    const size_t N = sq_hamiltonian.core().parameters(0).dimension(); // returns the number of electrons
```
lelemmen, line 122: Similarly, the right-hand side is not the number of electrons.

gqcp/tests/QCMethod/CC/CCD0_test.cpp
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    return F;
+}
+
+GQCP::BlockRankFourTensor<double> initializeAmplitudes(const GQCP::SQHamiltonian<double>& sq_hamiltonian, const GQCP::ScalarSQOneElectronOperator<double>& f){
+    const auto gs = sq_hamiltonian.twoElectron().parameters(0);
+    const auto ls = twoElectronAntiSymmetrized(gs);
+
+    const size_t N = gs.dimension(); // number of electrons
```
lelemmen, line 146: Idem.

gqcp/tests/QCMethod/CC/CCD0_test.cpp
```diff
@@ -0,0 +1,257 @@
...
+                }
+            }
+        }
+    }
+    return ts;
+}
+
+void step(const GQCP::SQHamiltonian<double>& sq_hamiltonian, GQCP::BlockRankFourTensor<double>& ts, GQCP::BlockRankFourTensor<double>& F, const GQCP::ScalarSQOneElectronOperator<double>& f){
+    const auto fs = f.parameters();
+    const size_t N = sq_hamiltonian.core().parameters(0).dimension(); // returns the number of electrons
```
lelemmen, line 174: Idem.

gqcp/tests/QCMethod/CC/CCD0_test.cpp
dariatols: Are you saying that the `twoElectronAntiSymmetrized` parameter should be a `SQTwoElectronOperator` instead of a `QCRankFourTensor` such that I can define `g_par` within the function?

dariatols: Meaning that the return type of the function should be a `SQTwoElectronOperator` too instead of a `QCRankFourTensor`?

gqcp/tests/QCMethod/CC/CCD0_test.cpp
dariatols: Foolish mistake indeed, copying large tensors could be costly 😁.

gqcp/tests/QCMethod/CC/CCD0_test.cpp
dariatols: Indeed, I was too fixated on the H2 problem with an STO-3G basis. The best way would probably be to pass `N` as a parameter. Later on, when I rearrange my code into classes, `N` could be a member variable then.

gqcp/tests/QCMethod/CC/CCD0_test.cpp
lelemmen: Yes, that's something I would suggest.

gqcp/tests/QCMethod/CC/CCD0_test.cpp
lelemmen: Indeed, in your current functional approach, I would also pass `N` is an argument. Later on, `N` should indeed be a member/property of some wrapping class.

dariatols:
@lelemmen Your issues are addressed in my latest commit. My energy is not converging well until this point, as it is always bouncing between 2 energies. I'm trying to resolve this issue and let you know if I've found my mistake.

lelemmen:
@dariatols I haven't looked into the exact implementation of the formulas, but could you explain why you're dividing indices by 2?

Also, you forgot to change the variable name `N` to `K` in your antisymmetrizing function.

dariatols:
With the adjustments of the loop indices, the final energy now converges to -1.1461, but should be 1.137306 hartree. I've double checked the `calculateProjection` function with the  [T^2 amplitude equations](http://www.esqc.org/lectures/CC_theory_2.pdf). Could this deviation be addressed to the fact that I've used a restricted basis set instead of a spinor basis set (as discussed in the call)? @lelemmen

lelemmen:
@dariatols Probably, yes. Could you run a psi4/pyscf calculation on the same system to check if your reference calculation is correct?

dariatols:
@lelemmen Using [this CCD modification from PySCF](https://github.com/pyscf/pyscf/blob/master/examples/cc/43-ccd.py), the converged energy for ```mol = gto.M(
    atom = 'H 0 0 -0.371000000; H 0 0 0.371000000',
    basis = 'STO-3G')``` becomes `E(CCD) = -1.13726350514568  E_corr = -0.02061225773415817`.

lelemmen:
@dariatols OK. Then I suppose something is going wrong in your current implementation. To be continued...

lelemmen:
@dariatols After PR #581, there's a lot of functionality added to GQCP concerning coupled-cluster. Maybe you could start off a new branch from `develop` (and discontinue this one), so you have access to the latest APIs.

If that would be too much of a hassle, feel free just to rebase off `develop`.

dariatols:
@lelemmen I'll start a new branch, since I'll be rewriting most of my code!

Issue #584, opened 2020-05-26
Implement spin expectation values for all HF models


lelemmen:
**Describe the feature you'd like**
It would be nice if the spin expectation values (for `S_x, S_y, S_z`) could be calculated from the HF `QCModel`s.

**Describe what the current code offers in relation to what it lacks**
It's already possible to calculate these spin expectation values generally, using a contraction of the quantized `ElectronicSpin` operator and the density matrix, so we can check this specialized implementation with the general implementation

**Additional context**

xdvriend:
I will tackle this issue in a separate PR, in order to prevent the current one from becoming too large.

The GHF expectation values are already implemented, so there's just RHF and UHF left to do.

xdvriend:
@lelemmen I'm a bit confused here. For UHF, I can implement the `S_z` expectation value using the traces of the alpha-alpha and beta-beta overlap matrices. Should the function just return 0 for `S_x` and `S_y`? As for RHF, I can do the same for `S_z`, but this essentially comes down to `trace - trace` which seems redundant to me. `For S_x` and `S_y` the same problem occurs s with UHF, since the mixed components vanish.

lelemmen:
Yep, for RHF and UHF you can hard-code the values wherever applicable.

xdvriend:
@lelemmen Is it possible that `RSpinorBasis` can't quantize an `ElectronicSpinOperator`, like the `GSpinorBasis` can?

lelemmen:
> @lelemmen Is it possible that `RSpinorBasis` can't quantize an `ElectronicSpinOperator`, like the `GSpinorBasis` can?

In principle, any spinor basis may quantize the electronic spin operator. However, in this code, we're making a difference between `RSpinorBasis` and `USpinorBasis`, versus `GSpinorBasis`. The first two are spin-orbital bases, representing the situation where the off-diagonal coefficient blocks in `GSpinorBasis` are zero. This means that one-electron integrals in the first two spin-orbital basis can be represented by `K x K`-matrices, while we use `2K x 2K`-matrices to represent the one-electron integrals in a `GSpinorBasis`.

As for your remark, I think that `RSpinorBasis` can quantize `ElectronicSpinOperator`, but the resulting `SQOneElectronOperator` should encapsulate `2K x 2K`-matrices, just like the the result of `GSpinorBasis`.

Your question seems related to #559, since the spin operators aren't singlet operators like the non-relativistic spin-free molecular Hamiltonian.

PR #586, opened 2020-05-26
GHF implementations

lelemmen:
**Short description**
In this PR, I'll implement GHF and related APIs. This PR will also implement
- [ ] GHF SCF DIIS
- [ ] Python bindings for GHF
- [ ] spin expectation values for RHF, UHF and GHF (#584)

**Related issues**

**Additional context**

**Don't forget:** 


lelemmen:
@xdvriend Thanks for the collaboration session we had earlier. I'm sure you learned a lot about how GQCP is structured, what objects are available, etc.

In fact, I think you're ready to start tackling some harder problems in GQCP. Would you be up for the implementation of GHF SCF DIIS? I think you have enough experience to be able to implement this.

If you're starting to feel comfortable (and you can also just take a look at the current RHF and UHF bindings), would you also be up for implementing the GHF Python bindings, and providing an example notebook with some reference GHF calculations?

xdvriend:
@lelemmen I definitely learned a lot, so thank you as well for putting the effort in these collaboration sessions.
I'd be happy to look into the things you've mentioned, provided I find some time to do so.

xdvriend:
Closed because of PR #638 .

PR #595, opened 2020-06-27
CCD0 bindings and implementation.

dariatols:
**Short description**
In this PR, I'll write bindings for CCD0 and use these to implement a CCD0 algorithm by means of an injection step.

**Don't forget:** 


dariatols:
There is still a bug at this Python binding.
```
.def("replace_current_t2_amplitudes",
            [](GQCP::CCSDEnvironment<double>& environment, const Eigen::Tensor<double, 4>& new_t2_amplitudes, const size_t N, const size_t M) {
                GQCP::T2Amplitudes<double> t2(GQCP::ImplicitRankFourTensorSlice<double>::FromBlockRanges(0, N, 0, N, 0, M, 0, M, new_t2_amplitudes), GQCP::OrbitalSpace::Implicit({{GQCP::OccupationType::k_occupied, N}, {GQCP::OccupationType::k_virtual, M - N}}));
                environment.t2_amplitudes.pop_back();
                environment.t2_amplitudes.push_back(t2);
             })
```
I think this is due to the wrong function argument `const Eigen::Tensor<double, 4>& new_t2_amplitudes`, but I have based myself on the bindings of `UHFSCFENvironment` where a similar argument `const Eigen::MatrixXd&` is given for a NumPy array.

dariatols:
> There is still a bug at this Python binding.
> 
> ```
> .def("replace_current_t2_amplitudes",
>             [](GQCP::CCSDEnvironment<double>& environment, const Eigen::Tensor<double, 4>& new_t2_amplitudes, const size_t N, const size_t M) {
>                 GQCP::T2Amplitudes<double> t2(GQCP::ImplicitRankFourTensorSlice<double>::FromBlockRanges(0, N, 0, N, 0, M, 0, M, new_t2_amplitudes), GQCP::OrbitalSpace::Implicit({{GQCP::OccupationType::k_occupied, N}, {GQCP::OccupationType::k_virtual, M - N}}));
>                 environment.t2_amplitudes.pop_back();
>                 environment.t2_amplitudes.push_back(t2);
>              })
> ```
> 
> I think this is due to the wrong function argument `const Eigen::Tensor<double, 4>& new_t2_amplitudes`, but I have based myself on the bindings of `UHFSCFENvironment` where a similar argument `const Eigen::MatrixXd&` is given for a NumPy array.

This issue is now fixed.

gqcpy/src/QCMethod/CC/CCSDEnvironment_bindings.cpp
```diff
@@ -28,6 +28,30 @@ namespace py = pybind11;
...
+    py::buffer_info buffer_info = inArray.request();
+
+    // extract data an shape of input array
+    T *data = static_cast<T *>(buffer_info.ptr);
+    std::vector<ssize_t> shape = buffer_info.shape;
+
+    // wrap ndarray in Eigen::Map
+    Eigen::Tensor<T, 4>  in_tensor = Eigen::TensorMap<Eigen::Tensor<T, 4>>(data, shape[0], shape[1], shape[2], shape[3]);
+    return in_tensor;
+}
```
lelemmen, line 52: Could you clean up this source code?
- The line comments don't start with a capital
- Sometimes, there are too many white spaces (use `clang-format` for this)
- There's no need to manually line break line 32
- You can use `const auto` in many cases.

gqcpy/src/QCMethod/CC/CCSDEnvironment_bindings.cpp
```diff
@@ -62,17 +86,35 @@ void bindCCSDEnvironment(py::module& module) {
...
+                const auto axis1_indices = orbital_space.indices(GQCP::OccupationType::k_occupied);
+                const auto axis2_indices = orbital_space.indices(GQCP::OccupationType::k_occupied);
+                const auto axis3_indices = orbital_space.indices(GQCP::OccupationType::k_virtual);
+                const auto axis4_indices = orbital_space.indices(GQCP::OccupationType::k_virtual);
+                
+                Eigen::Tensor<double, 4> t2_tensor = asTensor(new_t2_amplitudes);
+                GQCP::T2Amplitudes<double> t2{GQCP::ImplicitRankFourTensorSlice<double>::FromIndices(axis1_indices, axis2_indices, axis3_indices, axis4_indices, t2_tensor), orbital_space};
+                environment.t2_amplitudes.pop_back();
+                environment.t2_amplitudes.push_back(t2);
+            });
```
lelemmen, line 115: Could you clean up this file?
- Some white space is missing (use `clang-format`)
- Use `const`, `const auto`, `const` ref, ... wherever possible.

gqcpy/src/QCModel/CC/T2Amplitudes_bindings.cpp
```diff
@@ -26,6 +26,24 @@ namespace py = pybind11;
...
+py::array_t<T> asNumpyArray(const Eigen::Tensor<T, 4>& tensor) {
+
+    // Implementation adapted from https://github.com/pybind/pybind11/issues/1377
+    const auto shape = tensor.dimensions();
+
+    return py::array_t<T>(shape,
+                          {shape[0] * shape[1] * shape[2] * sizeof(T), shape[1] * shape[2] * sizeof(T), shape[2] * sizeof(T), sizeof(T)},  // strides
+                          tensor.data()                                                                                                    // data pointer
+    );
+}
```
lelemmen, line 46: This seems to be a copy-paste from another of our implementations. Please make sure that there is only one copy of this function available in the library.

dariatols: I copied since I saw no other way to implement this in a "clean" way in the include files. Or do you mean something else?

lelemmen: I might be mistaken, but couldn't you just create an "utility" header or something (not sure on the name), and `#include` it in both files?

lelemmen:
@dariatols Are the bugs you've reported concerning the non-converging T2 CCD amplitudes (with the injected 'print' functionality) resolved? If so, could you explain what you did to resolve them and what you think the issue was?

gqcpy/src/QCModel/CC/T2Amplitudes_bindings.cpp
dariatols: I'll add the `asNumpyArray` in `include/Utilities/Eigen.hpp` since it is an `Eigen`-functionality (in a way).

dariatols:
> @dariatols Are the bugs you've reported concerning the non-converging T2 CCD amplitudes (with the injected 'print' functionality) resolved? If so, could you explain what you did to resolve them and what you think the issue was?

@lelemmen This is still not solved. I think there is something wrong when I convert the `Eigen::Tensor` to a NumPy array since printing `np.any(t2_amplitudes>1e+08)` gives different results each time I restart the kernel.

gqcpy/src/QCModel/CC/T2Amplitudes_bindings.cpp
lelemmen: I think that would pollute the GQCP/gqcp core C++ library (we don't need this utility to build it), so I think it's better to add it inside `gqcpy/include`.

lelemmen:
I've checked the CCSD T2-amplitudes and I also find very large values (`np.any(t2_amplitudes>1e+08)`), so I'm not convinced that this is actually a problem, since these values may be contracted with antisymmetric integrals whose value is 0.

Maybe the convergence issues you've described are due to an incorrect implementation of the update of the T2-amplitudes in the CCD0-step?

dariatols:
But isn't it weird that np.any(t2_amplitudes>1e+08) changes each time you restart the kernel? @lelemmen

lelemmen:
I'm keeping this PR open, since it contains unmerged work.

PR #600, opened 2020-07-06
Magnetic inducibilities

lelemmen:
**Short description**
In this PR, the implementation of the magnetic inducibility will be addressed.

**Related issues**

**Additional context**

**Don't forget:**


lelemmen:
I've just noticed that the perturbation integrals (angular and linear momentum) have not been transformed from the AO to the canonical RHF basis.

- [ ] fix canonical momentum integrals

PR #637, opened 2020-08-25
Website improvements

lelemmen:
**Short description**
Here are the changes that we did to our website in our most recent live share session.

**Related issues**

**Additional context**

**Don't forget:**


xdvriend:
On the current homepage of our website, there is still a lot of placeholder information. However, I personally do quite like the division in different block, as it is currently. 

So with that in mind, I wanted to ask everyone (@guacke @lelemmen @dariatols @nbilliet ) What sections you would like to see on the homepage. The current introduction + features in the beginning and academic references at the bottom would stay as they are now, everything in between can change.

I thought maybe a section about our group would be nice? All suggestions are welcome!

dariatols:
@xdvriend A section about our group would be nice. We could refer to the members page in `gqcg-org/GQCG`?

dariatols (approved):

xdvriend:
The last commit should solve issue #656 .

PR #654, opened 2020-09-01
Add support for GQCP Docker images.

### LINKED ISSUE #549: Unknown exception errors when calculating integrals through Python bindings
**Describe the bug**
Installation on Google Colab leads to 'unknown exception' errors when calculating integrals through the Python bindings.

**To Reproduce**
```python

import gqcpy
import numpy as np
import matplotlib.pyplot as plt

def gqcp_hydrogen_chain(num_hydrogens, distance):
    nuclei = []
    for index in range(num_hydrogens):
        nuclei.append(gqcpy.Nucleus(1, 0.0, 0.0, index*distance))
    molecule = gqcpy.Molecule(nuclei)
    return molecule

molecule = gqcp_hydrogen_chain(4, 4)
spinor_basis = gqcpy.RSpinorBasis(molecule = molecule, basisset_name="sto-3g")
```

**Proposed solution**
This exception can occur because the Python bindings are currently tested for import; not for functionality. A Python test that checks if integrals can be calculated will allow us to resolve this problem upstream.

### LINKED ISSUE #652: GQCP Docker image
Make a Docker image for every develop release.

**Additional context**
Also provide documentation for how to run remote Jupyter notebooks.

guacke:
**Short description**
This PR will add support for running GQCPy in docker images. It:
- closes #652;
- closes #549;

**Don't forget:**
- [x] This PR should also add documentation on how to run Jupyter notebooks on the HPC


guacke:
Problem importing GQCPY:

```bash
Python 3.8.3 (default, May 19 2020, 18:47:26) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import gqcpy
Illegal instruction
```
Will add unit test that catches this behavior before deployment.

guacke:
Is related to issue #549.

guacke:
The problem I mention above has essentially _nothing_ to do with GQCP and _everything_ to do with GQCPy. This hints that it could be beneficial to split GQCP from GQCPy (as is also discussed in the final paragraph of [1]):
- It makes 'sense' to have both a GQCP and a GQCPy repo with separate CI/CD as both builds are concerned with other issues (building a C++ library vs building Python bindings). 
- This is even reflected in the current directory structure of this codebase: gqcp is a separate directory from gqcpy. 
- For a 'typical' C++ library, such as e.g. [2], one would expect the include, src and test directory to be at the top-level. 
- The 'job' of the GQCP repo is then to develop the C++ library; the 'job' of the GQCPy repo is to open up those routines to Python interfaces. 
- The testing suites of both repos should be testing very different aspects of the resulting codebases. 
- The GQCPy repo could gather a set of Jupyter notebook tutorials, while this seems more artificial in the C++ GQCP repo.

@lelemmen What are your thoughts?

[1] https://www.benjack.io/2018/02/02/python-cpp-revisited.html
[2] https://github.com/boostorg/math

lelemmen:
I'm neither really against nor really for splitting up the C++ library from its bindings. I agree that there are two aspects of this software package, i.e. the C++ code and generating its Python bindings, but should that mean that they should be split up? (Should we also split up the source code for the docusaurus website?) What exactly is the benefit of splitting them up? What is possible in the split-up case that isn't in the current situation?

As for some counter-thoughts, I think the following issues (pun intended) should at least be tackled before initiating this split:
- How can the project management be efficiently done using GitHub, for two separate but related repositories?
- How do we keep the versioning in sync? (We currently still don't really have a stable API, so any small refactor would cause the Python bindings to be incompatible again.)

guacke:
This PR enables support for Docker `GQCP`. I have also taken the opportunity to refactor the Github Actions workflows and add a test for the Python bindings. As such, this PR closes some additional issues that were related to those bindings.

.github/workflows/website.yml
```diff
@@ -50,11 +53,20 @@ jobs:
...
+      if: github.event_name == 'pull_request'
+    
     # Deploy the website to GitHub pages.
     - name: Deploy
       uses: peaceiris/actions-gh-pages@v3
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         publish_dir: ./site/website/build/GQCP
         publish_branch: gh-pages
+      if: github.event_name == 'push'
```
lelemmen, line 72: Does this mean that the website is only updated on a push to develop?

gqcpy/test_bindings.py
```diff
@@ -0,0 +1,2 @@
+import gqcpy
```
lelemmen, line 1: Could you put this test in the `gqcpy/tests/` folder?

lelemmen (approved):
Thanks for all the work on providing Docker for GQCP, and providing documentation on how to use it in conjunction with Jupyter Notebooks!

.github/workflows/website.yml
guacke: Yes

gqcpy/test_bindings.py
guacke: Addressed in latest commit.

guacke:
Apparently, `-march=native` is a no-go when using Docker [1]. This will cause the GQCP library inside the Docker image to use the optimisations offered by the local build machine, which can cause errors when these optimisations are not available on the run machine. I am working on resolving this issue as this currently makes running GQCP Singularity on Golett impossible due to 'Illegal instructions'.

[1] https://stackoverflow.com/questions/54039176/mtune-and-march-when-compiling-in-a-docker-image

guacke:
The `march=native` compiler flag was erroneously set to public in our `Libcint` port. Fixing this in the Conda package however led to interoperability issues with our `libint` conda package. As such, our current Conda CD is **broken**. In order to fix this, I first have to enable CI/CD for our Libint 2.3.1 package. This should fix our conda installation; afterwards I will return to this PR and check that `march=native` is no longer being used.

guacke:
Previous commit switches to conda-forge Libint which seems to work. Will investigate further on why this is the case in the gqcg-oss repo as conda-forge does not supply Libint 2.4.2 MacOS packages.

guacke:
@lelemmen Disabling test BOOST_AUTO_TEST_CASE(One_DM_from_2_DM) due to failing test.

guacke:
@lelemmen @dariatols @xdvriend This PR re-enables CI/CD for **linux** (including Docker/Singularity). The MacOS build is currently stuck due to errors with Libint conda. I will resolve this in a different PR. After merge with develop, all outstanding PRs should rebase on develop for CI/CD to work again on those branches.

site/docs/developer_documentation/devops.md
```diff
@@ -0,0 +1,40 @@
+---
+id: devops
+title: GQCP Devops
```
lelemmen, line 3: What is the preferred capitalization? Is it "DevOps" or "Devops"?

lelemmen (approved):
Ugh. What a mess this must have been to fix this these last couple of days! Thanks for bringing us one step closers to the clouds!

.github/workflows/conda.yml
```diff
@@ -18,7 +18,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        #os: [ubuntu-latest, macos-latest]
```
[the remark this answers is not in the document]
guacke: @lelemmen I like your cloud-oriented spirit. However, I would still like to support MacOS installations that are closer to the metal than Docker when running on Mac (mainly aimed at our own server infrastructure). As described above, MacOS support will be fixed in a separate PR.

CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ option(BUILD_TESTS "Build the C++ library tests" OFF)
 option(BUILD_BENCHMARKS "Build the executables used for benchmarking the C++ library" OFF)
 option(BUILD_DOCS "Build the documentation of the C++ library using Doxygen" OFF)
 option(BUILD_PYTHON_BINDINGS "Build the Python bindings for the C++ library" OFF)
+option(OPTIMIZE_FOR_NATIVE "Build with -march=native" OFF)
```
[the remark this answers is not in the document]
guacke: Will be added in last commit.

gqcp/tests/Processing/DensityMatrices/CIDMCalculators/SpinResolvedDMCalculator_test.cpp
```diff
@@ -115,7 +117,7 @@ BOOST_AUTO_TEST_CASE(One_DM_from_2_DM) {
     const auto D_from_reduction = (1.0 / (N - 1)) * d.reduce();
     BOOST_CHECK(D_from_reduction.isApprox(D, 1.0e-12));
 }
-
+*/
```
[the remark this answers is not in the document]
guacke: Note that this can be related to disabling the `march=native` optimization flags. I will open an issue that is aimed at documenting this issue.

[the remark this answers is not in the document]
guacke: This is issue #667.

site/docs/developer_documentation/devops.md
guacke: I prefer DevOps. The id is in lowercase.

PR #671, opened 2020-09-08
Use unresolved and unrestricted types in UHF module

### LINKED ISSUE #651: Use spin-resolved types in the UHF module
Currently, we're still using two density matrices to represent the alpha and beta part of the density matrix.

We have introduced spin-resolved density matrices in previous PRs, so we should use them!

xdvriend:
**Short description**
Changing the UHF module to make full use of the SpinResolved1DM and USQOperators.

**Related issues**
Closes #651 

**Don't forget:**


gqcp/include/QCMethod/HF/UHF/UHFFockMatrixCalculation.hpp
```diff
@@ -62,9 +62,10 @@ class UHFFockMatrixCalculation:
 
         const auto& P_alpha = environment.density_matrices_alpha.back();  // the most recent alpha density matrix
         const auto& P_beta = environment.density_matrices_beta.back();    // the most recent beta density matrices
+        SpinResolvedOneDM<Scalar> P {P_alpha, P_beta};
 
-        const auto F_alpha = QCModel::UHF<double>::calculateScalarBasisFockMatrix(Spin::alpha, P_alpha, P_beta, environment.sq_hamiltonian);
-        const auto F_beta = QCModel::UHF<double>::calculateScalarBasisFockMatrix(Spin::beta, P_alpha, P_beta, environment.sq_hamiltonian);
+        const auto F_alpha = QCModel::UHF<double>::calculateScalarBasisFockMatrix(Spin::alpha, P, environment.sq_hamiltonian);
+        const auto F_beta = QCModel::UHF<double>::calculateScalarBasisFockMatrix(Spin::beta, P, environment.sq_hamiltonian);
```
lelemmen, line 68: The method `QCModel::UHF::calculateScalarBasisFockMatrix` could also return an `USQOneElectronOperator`.

gqcp/include/QCModel/HF/UHF.hpp
```diff
@@ -171,12 +173,14 @@ class UHF {
...
+     *  @param N_a          the number of alpha electrons, i.e. the number of occupied alpha spin-orbitals
+     *  @param N_b          the number of beta electrons, i.e. the number of occupied beta spin-orbitals
      *
-     *  @return the sigma-spin UHF 1-DM expressed in an orthonormal sigma spin-orbital basis
+     *  @return the SpinResolved UHF 1-DM expressed in an orthonormal sigma spin-orbital basis
      */
-    static OneDM<Scalar> calculateOrthonormalBasis1DM(const size_t K_sigma, const size_t N_sigma) {
+    static SpinResolvedOneDM<Scalar> calculateOrthonormalBasis1DM(const size_t K_a, const size_t K_b, const size_t N_a, const size_t N_b) {
 
         // The 1-DM for UHF looks like (for K=5, N=3)
```
lelemmen, line 185:
```suggestion
        // The (alpha) 1-DM for UHF looks like (for K_alpha=5, N_alpha=3)
```

gqcp/include/QCModel/HF/UHF.hpp
```diff
@@ -198,13 +207,19 @@ class UHF {
      *
      *  @return the sigma-spin UHF 1-DM expressed in the underlying scalar basis
```
lelemmen, line 208: Please update the documentation accordingly.

gqcp/include/QCModel/HF/UHF.hpp
```diff
@@ -198,13 +207,19 @@ class UHF {
...
-        const auto K_sigma = C_sigma.numberOfOrbitals();
-        const auto D_orthonormal = UHF<Scalar>::calculateOrthonormalBasis1DM(K_sigma, N_sigma);
+        const auto K_a = C_a.numberOfOrbitals();
+        const auto K_b = C_b.numberOfOrbitals();
+        const auto D_orthonormal = UHF<Scalar>::calculateOrthonormalBasis1DM(K_a, K_b, N_a, N_b);
+
+        // Calculate the alpha and beta scalar basis 1 DM
+        OneDM<Scalar> D_AO_a = C_a.conjugate() * D_orthonormal.alpha() * C_a.transpose();
+        OneDM<Scalar> D_AO_b = C_b.conjugate() * D_orthonormal.beta() * C_b.transpose();
+        SpinResolvedOneDM<Scalar> D_AO {D_AO_a, D_AO_b};
```
lelemmen, line 219: Instead of manually transforming a spin-resolved 1-DM to the AO basis, could you add this API into `SpinResolvedOneDM`?

gqcp/include/QCModel/HF/UHF.hpp
```diff
@@ -217,11 +232,11 @@ class UHF {
      * 
      *  @return the UHF direct (Coulomb) matrix for spin sigma
      */
-    static ScalarSQOneElectronOperator<Scalar> calculateScalarBasisDirectMatrix(const OneDM<Scalar>& P_alpha, const OneDM<Scalar>& P_beta, const SQHamiltonian<Scalar>& sq_hamiltonian) {
+    static ScalarSQOneElectronOperator<Scalar> calculateScalarBasisDirectMatrix(const SpinResolvedOneDM<Scalar>& P, const SQHamiltonian<Scalar>& sq_hamiltonian) {
```
lelemmen, line 235: This method should return a `ScalarUSQoneElectronOperator`, but the other refactor is good!

gqcp/include/QCModel/HF/UHF.hpp
```diff
@@ -306,10 +319,22 @@ class UHF {
      */
     OneDM<Scalar> calculateOrthonormalBasis1DM(const Spin sigma) const {
```
lelemmen, line 320: This method should return a `SpinResolvedOneDM`.

gqcp/include/QCModel/HF/UHF.hpp
```diff
@@ -320,10 +345,22 @@ class UHF {
      */
     OneDM<Scalar> calculateScalarBasis1DM(const Spin sigma) const {
```
lelemmen, line 346: This method should return a `SpinResolvedOneDM`.

lelemmen (commented):
This PR is headed into the correct direction! While the refactor that you've already performed is correct, we could benefit from driving it even further.

gqcp/include/QCModel/HF/UHF.hpp
xdvriend: I thought this API was meant to return either the `alpha` or `beta` component of the `OrthonormalBasis1DM` (`SpinResolved1DM`). If this is not the case, should I add an API for that as well?

gqcp/include/QCModel/HF/UHF.hpp
xdvriend: Same question as for the `calculateOrthonormalBasis1DM` API.

gqcp/include/QCModel/HF/UHF.hpp
lelemmen: You can remove the 'old' functionality that was meant to calculate only the alpha or beta part, and replace it with the new functionality that calculates them both and encapsulates them correctly.

gqcp/include/QCModel/HF/UHF.hpp
lelemmen: Same answer, here. You can remove the 'old' functionality that was meant to calculate only the alpha or beta part, and replace it with the new functionality that calculates them both and encapsulates them correctly.

xdvriend:
To do: Adjust OneDM and SpinResolvedOneDM transformation functions.

xdvriend:
I've encountered an issue where I can't get the UHF algorithm to converge at the moment. I believe it has something to do with the convergence criterion, and the way it is constructed. I'll keep looking into it and I'll keep updating this PR to show my work process.

gqcp/include/QCMethod/HF/UHF/UHFSCFSolver.hpp
```diff
@@ -69,16 +69,10 @@ class UHFSCFSolver {
...
-        const SingleConvergenceType convergence_criterion_alpha {threshold, density_matrix_alpha_extractor, "the UHF alpha-density matrix in AO basis"};
-
-        const auto density_matrix_beta_extractor = [](const UHFSCFEnvironment<Scalar>& environment) { return environment.density_matrices_beta; };
-        const SingleConvergenceType convergence_criterion_beta {threshold, density_matrix_beta_extractor, "the UHF beta-density matrix in AO basis"};
-
-        const CompoundConvergenceCriterion<UHFSCFEnvironment<Scalar>> convergence_criterion {convergence_criterion_alpha, convergence_criterion_beta};
+        using SingleConvergenceType = ConsecutiveIteratesNormConvergence<SpinResolvedOneDM<Scalar>, UHFSCFEnvironment<Scalar>>;
 
+        const auto density_matrix_extractor = [](const UHFSCFEnvironment<Scalar>& environment) { return environment.density_matrices; };
+        const SingleConvergenceType convergence_criterion {threshold, density_matrix_extractor, "the UHF density matrices in AO basis"};
```
lelemmen, line 75: I see that you're now using the `.norm`-method that you've implemented on `SpinResolvedOneDM, which is the reason you're reporting convergence problems.

gqcp/include/Processing/DensityMatrices/SpinResolvedOneDM.hpp
```diff
@@ -89,6 +89,13 @@ class SpinResolvedOneDM {
      */
     const OneDM<Scalar>& beta() const { return this->D_bb; }
 
+    /**
+     *  @return the sum of the norms of the alpha and beta components of the spin resolved one DM
+     */
+    size_t norm() const {
+        return this->alpha().norm() + this->beta().norm();
```
lelemmen, line 96: You indeed return the sum of the alpha- and beta-norms, but is this really the Frobenius norm of the total, spin-resolved 1-DM?

gqcp/include/QCMethod/HF/UHF/UHFFockMatrixDIIS.hpp
```diff
@@ -97,22 +97,28 @@ class UHFFockMatrixDIIS:
...
+            beta_fock_matrices.push_back(fock_matrices[i].parameters(Spin::beta));
+        }
 
         // Calculate the accelerated Fock matrices and do a diagonalization step on them.
-        const auto F_alpha_accelerated = this->diis.accelerate(fock_matrices_alpha, error_vectors_alpha);
-        const auto F_beta_accelerated = this->diis.accelerate(fock_matrices_beta, error_vectors_beta);
+        const auto F_alpha_accelerated = this->diis.accelerate(alpha_fock_matrices, error_vectors_alpha);
+        const auto F_beta_accelerated = this->diis.accelerate(beta_fock_matrices, error_vectors_alpha);
+
+        const ScalarUSQOneElectronOperator<Scalar> Fock_accelerated {F_alpha_accelerated, F_beta_accelerated};
```
lelemmen, line 114:
```suggestion
        const ScalarUSQOneElectronOperator<Scalar> F_accelerated {F_alpha_accelerated, F_beta_accelerated};
```

gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp
```diff
@@ -88,21 +88,49 @@ void bindUHFSCFEnvironment(py::module& module) {
...
 
-        .def_readonly(
+        .def(
             "density_matrices_alpha",
-            &GQCP::UHFSCFEnvironment<double>::density_matrices_alpha)
+            [](GQCP::UHFSCFEnvironment<double>& environment) {
+                std::vector<GQCP::MatrixX<double>> alpha_density_matrices;
 
-        .def_readonly(
+                for (int i = 0; i < environment.density_matrices.size(); i++) {
```
lelemmen, line 96: It's best that you use `size_t` for such a loop.

gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp
```diff
@@ -88,21 +88,49 @@ void bindUHFSCFEnvironment(py::module& module) {
             "coefficient_matrices_beta",
             &GQCP::UHFSCFEnvironment<double>::coefficient_matrices_beta)
 
-        .def_readonly(
+        .def(
```
lelemmen, line 91: Could you explain why you changed from `.def_readonly` to `.def`?

gqcp/include/QCMethod/HF/UHF/UHFSCFSolver.hpp
xdvriend: Yeah, I already figured something was going wrong there. I was looking into it already. I tried to implement it more similar to how it first was, with a separate lambda function to extract the alpha and beta density matrices, but that didn't work either.

gqcp/include/Processing/DensityMatrices/SpinResolvedOneDM.hpp
xdvriend: I'll look into correcting this function, hopefully that fixes the issues I've been having!

gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp
xdvriend: When using `def_readonly()` I get an error while compiling gqcpy:

`no matching member function for call to 'def_readonly'`
`candidate template ignored: could not match 'const D C::*' against '(lambda at /Users/xeno/Documents/GQCP/gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp:93:13)'`

I wanted to ask you about this, since I wasn't sure about changing the `def_readonly()` bindings to regular `def()` ones myself. But I was first trying to fix the convergence problems.

gqcp/include/QCMethod/HF/UHF/UHFSCFSolver.hpp
xdvriend: I discovered some additional info while trying to resolve this problem. 

-  When using the code above, with the (wrong) `.norm()` function, the method does converge, but it converges wrongly. (I think it stops too early, leading to an energy estimate that's not low enough). 

- As mentioned, I also tried another method, more similarly implemented to the original code. However this causes the algorithm to stop converging altogether. I've added the code below in order to show you my thought process.

```C++
const auto density_matrix_alpha_extractor = [](const UHFSCFEnvironment<Scalar>& environment) {
            std::deque<GQCP::OneDM<Scalar>> alpha_density_matrices;

            for (size_t i; i < environment.density_matrices.size(); i++) {
                alpha_density_matrices.push_back(environment.density_matrices[i].alpha());
            }
            return alpha_density_matrices;
        };

        const auto density_matrix_beta_extractor = [](const UHFSCFEnvironment<Scalar>& environment) {
            std::deque<GQCP::OneDM<Scalar>> beta_density_matrices;

            for (size_t i; i < environment.density_matrices.size(); i++) {
                beta_density_matrices.push_back(environment.density_matrices[i].beta());
            }
            return beta_density_matrices;
        };

        const SingleConvergenceType convergence_criterion_alpha {threshold, density_matrix_alpha_extractor, "the UHF alpha-density matrix in AO basis"};
        const SingleConvergenceType convergence_criterion_beta {threshold, density_matrix_beta_extractor, "the UHF beta-density matrix in AO basis"};
        const CompoundConvergenceCriterion<UHFSCFEnvironment<Scalar>> convergence_criterion {convergence_criterion_alpha, convergence_criterion_beta};
```

gqcp/include/Processing/DensityMatrices/SpinResolvedOneDM.hpp
xdvriend: I tried to solve it by creating the generalised density matrix from the alpha and beta components.
```C++
size_t norm() const {

        const auto dim = this->alpha().numberOfOrbitals();
        OneDM<double> generalized_density = OneDM<double>::Zero(dim * 2, dim * 2);

        generalized_density.topLeftCorner(dim, dim) = this->alpha();
        generalized_density.bottomRightCorner(dim, dim) = this->beta();

        return generalized_density.norm();
    }
```
 The result however is exactly the same as with the old function shown above.

gqcp/include/QCMethod/HF/UHF/UHFSCFSolver.hpp
xdvriend: EDIT 2: Disccovered the mistake in my loops and this method should work fine now.

gqcp/include/Processing/DensityMatrices/SpinResolvedOneDM.hpp
xdvriend: Since I changed the implementation method, I removed this API for now.

xdvriend:
There is still something that's not working with the DIIS algorithm. I'm working on resolving this now.

xdvriend:
I think everything is adjusted now, So I have re-requested your review @lelemmen. I have left one conversation open, since I am not sure what to do with the python bindings that went from `def_readonly()` to `def()`. I am, as always open to suggestions about this, as well as the other adjustments I made to the code.

gqcp/include/Processing/DensityMatrices/OneDM.hpp
```diff
@@ -44,6 +44,17 @@ class OneDM: public QCMatrix<_Scalar> {
...
+    /*
+     * PUBLIC METHODS
+     */
+    
+    /**
+     *  @param T          transformation matrix for the spin unresolved 1-DM
+     * 
+     *  @return the transformed density matrix.
+     */
+    OneDM<Scalar> transform(const TransformationMatrix<double>& T) const { return this->basisTransform(); }
```
lelemmen, line 57:
```suggestion
    OneDM<Scalar> transformed(const TransformationMatrix<double>& T) const { return this->basisTransform(T); }
```

I guess you forgot an argument here. Also, I'd like to suggest the name `transformed` instead of `transform`, indicating that [a modification is being returned to the caller](https://swift.org/documentation/api-design-guidelines/).

gqcp/include/Processing/DensityMatrices/SpinResolvedOneDM.hpp
```diff
@@ -128,14 +121,25 @@ class SpinResolvedOneDM {
...
+     *  @param T_a          transformation matrix for the alpha component of the spin resolved 1-DM
+     *  @param T_b          transformation matrix for the beta component of the spin resolved 1-DM
+     * 
+     *  @return the transformed spin resolved density matrix, with each component transformed seperately t a different basis.
      */
-    SpinResolvedOneDM<Scalar> transformToScalarBasis(const TransformationMatrix<double>& C_a, const TransformationMatrix<double>& C_b) const {
-        OneDM<Scalar> D_AO_a = C_a.conjugate() * this->alpha() * C_a.transpose();
-        OneDM<Scalar> D_AO_b = C_b.conjugate() * this->beta() * C_b.transpose();
-        SpinResolvedOneDM<Scalar> D_AO {D_AO_a, D_AO_b};
+    SpinResolvedOneDM<Scalar> transform(const TransformationMatrix<double>& T_a, const TransformationMatrix<double>& T_b) const {
```
lelemmen, line 129:
```suggestion
    SpinResolvedOneDM<Scalar> transformed(const TransformationMatrix<double>& T_a, const TransformationMatrix<double>& T_b) const {
```

gqcp/include/Processing/DensityMatrices/SpinResolvedOneDM.hpp
```diff
@@ -128,14 +121,25 @@ class SpinResolvedOneDM {
...
+    }
+
+    /**
+     *  @param T          transformation matrix for the alpha and beta component of the spin resolved 1-DM
+     * 
+     *  @return the transformed spin resolved density matrix, with each component transformed to the same basis.
+     */
+    SpinResolvedOneDM<Scalar> transform(const TransformationMatrix<double>& T) const {
+        return this->transform(T, T);
     }
```
lelemmen: Thanks for implementing these methods!

Since you're providing two `TransformationMatrix` arguments, I would propose that we can take the spin-resolved types even further. Can you implement the (generic) type `SpinResolvedTransformationMatrix`, which effectively encapsulates `T_alpha` and `T_beta`? I think the extra encapsulation will be useful in later code! As for the public methods on `SpinResolvedTransformationMatrix`:
- `alpha()`
- `beta()`
- `transformed()`
- `transform()`

These are quite similar to `TransformationMatrix` and other spin-resolved types. Spin-resolved types should then also obtain the API to be transformed with a `SpinResolvedTransformationMatrix`.

gqcp/include/QCMethod/HF/UHF/UHFSCFSolver.hpp
```diff
@@ -69,10 +69,29 @@ class UHFSCFSolver {
...
+
+            for (size_t i = 0; i < environment.density_matrices.size(); i++) {
+                beta_density_matrices.push_back(environment.density_matrices[i].beta());
+            }
+            return beta_density_matrices;
+        };
+
+        const SingleConvergenceType convergence_criterion_alpha {threshold, density_matrix_alpha_extractor, "the UHF alpha-density matrix in AO basis"};
+        const SingleConvergenceType convergence_criterion_beta {threshold, density_matrix_beta_extractor, "the UHF beta-density matrix in AO basis"};
+        const CompoundConvergenceCriterion<UHFSCFEnvironment<Scalar>> convergence_criterion {convergence_criterion_alpha, convergence_criterion_beta};
```
lelemmen: I understand that this code makes the UHF algorithms re-converge. However, I'll take a little time to check why the convergence didn't work when using the spin-blocked (M x M) density matrix.

gqcpy/src/QCModel/HF/UHF_bindings.cpp
```diff
@@ -34,19 +34,17 @@ void bindQCModelUHF(py::module& module) {
...
         .def(
             "calculateScalarBasis1DM",
-            [](const GQCP::QCModel::UHF<double>& uhf_parameters, const GQCP::Spin sigma) {
-                return uhf_parameters.calculateScalarBasis1DM(sigma);
+            [](const GQCP::QCModel::UHF<double>& uhf_parameters) {
+                return uhf_parameters.calculateScalarBasis1DM();
             },
-            py::arg("sigma"),
-            "Return the sigma-spin UHF 1-DM expressed in the underlying scalar basis for these UHF model parameters.")
+            "Return the spin resolved UHF 1-DM expressed in the underlying scalar basis for these UHF model parameters.")
```
lelemmen, line 47: Since the Python bindings don't have the density matrices yet, any Python calls to these APIs will result in an error. Therefore, I'd like to propose for you to work on #677 after this one.

lelemmen (commented):
I guess this PR is almost done. 👍 Can we iterate once more on the spin-resolved types?

gqcp/include/Processing/DensityMatrices/SpinResolvedOneDM.hpp
xdvriend: I have made an issue for this, to which I assigned myself. That way, I can tackle this in a separate PR.

gqcpy/src/QCModel/HF/UHF_bindings.cpp
xdvriend: I will implement these bindings (from issue #677) in this PR, in order to make the UHF bindings work.

lelemmen:
Today I'll be able to check the convergence issues. Afterwards, we'll merge this PR.

gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp
lelemmen: I looked it up in the [documentation](https://pybind11.readthedocs.io/en/stable/classes.html#instance-and-static-fields). Did you perhaps look over this section? `.def_readonly` can only be used to bind a C++ class' _member_, which is not what we're trying to accomplish here. However, once the Python bindings contain classes for the appropriate return types, (probably in #677), it will be possible to re-write this using the previous `.def_readonly` implementation.

Also, in these density/Fock matrix bindings, you've written code to transform a `deque` to a `vector`. While I appreciate the effort of manually writing this loop, there are a couple of improvements that can be made:
- Lose the C-style loop, and rewrite it using `for (const auto& density_matrix : environment.density_matrices)`;
- Or, and this is preferred, use the C++ STL's [`std::transform`](https://en.cppreference.com/w/cpp/algorithm/transform) method to modify C++'s native types.

xdvriend: I must indeed have looked over this section, because I remember spending quite some time looking up why `def_readonly()` did no longer work. But thanks for clarifying! I'll look into rewriting the loops a swell.

gqcp/include/Processing/DensityMatrices/SpinResolvedOneDM.hpp
lelemmen: Erm... Could it be that you're failing to converge because you're returning `size_t` instead of double?

xdvriend: Oh wow, yup, that'll probably be it. Ill change everything accordingly and run the tests. If everything works I'll request your review on this PR one final time.

xdvriend:
@lelemmen , I believe the only unresolved thing in this PR are the python bindings that, at the moment, can't return anything because the density matrices don't yet have python bindings. Should I leave them in for now & tackle #677 next to resolve that? Or should I maybe comment them out and activate them again in the PR for the density matrix bindings?

If I can leave it as it is now, I believe everything in this PR is resolved.

lelemmen (approved):

Issue #663, opened 2020-09-04
Spin-redistribution implementations


lelemmen:
An epic for tracking progress on the implementation of the prerequisites for spin-redistribution implementations.

xdvriend:
@ lelemmen Is it possible we need the ability to calculate spin density matrices to be able to perform spin constrained CI calculations?

xdvriend:
Issue #680 ties into this issue, as it is needed to further the implementation of these API's.

Issue #680, opened 2020-09-15
Calculate mixed Coulomb operator with USpinorBasis


xdvriend:
**Describe the feature you'd like**
In order to rewrite the USQHamiltonian as cleanly as possible, the USpinorBasis needs an API that can calculate mixed two electron terms.

**Describe what the current code offers in relation to what it lacks**
At the moment, there is only the possibility to calculate pure spin coulomb terms, as can be seen here:

```C++
/**
  *  @param fq_op                the first-quantized Coulomb operator
  *  @param sigma                alpha or beta
  * 
  *  @return the second-quantized operator corresponding to the Coulomb operator in the spinor basis of the requested component
  * 
  *  @note This method is not (yet) capable of calculating 'mixed' integrals such as g_aabb.
  */
    auto quantize(const CoulombRepulsionOperator& fq_op, const Spin& sigma) const -> SQTwoElectronOperator<product_t<CoulombRepulsionOperator::Scalar, ExpansionScalar>, CoulombRepulsionOperator::Components> { return this->spinor_bases[sigma].quantize(fq_op); }
```

PR #684, opened 2020-09-17
USQhamiltonian refactor

xdvriend:
**Short description**
This PR aims to refactor the USQ Hamiltonian in order to progress the other implementations needed for the constrained CI research.

**Related issues**
- #663 
- #558 

**Additional context**
Things to look into:

- [ ] SelectedONVBasis tests fail at the moment. 
- [x] Something still malfunctions with single component Jacobi rotation (test commented out at the moment) (RESOLVED)

**Don't forget:**

### DIFF
--- a/gqcp/include/Basis/transform.hpp
+++ b/gqcp/include/Basis/transform.hpp
@@ -187,10 +187,10 @@ void basisRotate(USpinorBasis<TransformationScalar, Shell>& spinor_basis, USQHam
  *  @param component                        the spin component
  */
 template <typename Shell>
-void basisRotate(USpinorBasis<double, Shell>& spinor_basis, USQHamiltonian<double>& sq_hamiltonian, const JacobiRotationParameters& jacobi_rotation_parameters, const Spin& component) {
+void basisRotate(USpinorBasis<double, Shell>& spinor_basis, USQHamiltonian<double>& usq_hamiltonian, const JacobiRotationParameters& jacobi_rotation_parameters, const Spin& component) {
 
     spinor_basis.rotate(jacobi_rotation_parameters, component);
-    sq_hamiltonian.rotate(jacobi_rotation_parameters, component);
+    usq_hamiltonian.rotate(jacobi_rotation_parameters, component);
 }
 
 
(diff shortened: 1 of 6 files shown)


xdvriend:
The old implementation saved one SQTwoElectronOperator in the USQHamiltonian for the mixed two electron components. Initially, I will keep this by setting g_ba to zero, as all other implementations are based on this. Once the tests work, I can adapt this as well by first addressing issue #680 And refactoring the rest of the code with this in mind.

xdvriend:
- `SpinResolvedONVBasis_test` fails because `selected_fock_space.evaluateOperatorDiagonal(usq_hamiltonian)` finds the wrong solution
- `SpinResolvedFrozenONVBasis_test`fails because both the diagonals and hamiltonian evaluations are calculated wrongly.

xdvriend:
- There appears to be something wrong with `evaluateOperatorDiagonal(USQHamiltonian)` from the `SpinResolvedSelectedONVBasis`
- There appears to be a problem with `evaluateOperatorMatrixVectorProduct(USQHamiltonian)` from the `SpinResolvedSelectedONVBasis`

xdvriend:
In the function `evaluateOperatorDiagonal(USQHamiltonian)` in the `SpinResolvedSelectedONVBasis` you extract 5 parameters from the `USQHamiltonian`:
```C++
const auto& h_a = usq_hamiltonian.spinHamiltonian(Spin::alpha).core().parameters();
const auto& g_a = usq_hamiltonian.spinHamiltonian(Spin::alpha).twoElectron().parameters();
const auto& h_b = usq_hamiltonian.spinHamiltonian(Spin::beta).core().parameters();
const auto& g_b = usq_hamiltonian.spinHamiltonian(Spin::beta).twoElectron().parameters();

// Only g_ab is stored, for integrals derived from g_ba we reverse the indices as follows : g_ab(pqrs) = g_ba(rspq)
const auto& g_ab = usq_hamiltonian.twoElectronMixed().parameters();
```

I have checked all these parameters versus the old implementation of the `USQHamiltonian`, and they are all the same. (I used the old implementation to extract these values as well and put them in a `BOOST` test, to check whether they were equal to what I got with my implementation using the `isApprox()` function.)

However, the test still fails. I really don't see how the function can find a different result, while all the used parameters are the same. Do you have an idea where this problem can possibly originate @lelemmen ?

PR #692, opened 2020-09-25
Density matrix bindings

### LINKED ISSUE #677: Density matrix python bindings
In order to avoid adding unnecessary API's to certain objects, it would be beneficial to add python bindings for our density matrix API's.

Currently, these bindings are not part of the code.

xdvriend:
**Short description**
The functionality to work with both spin resolved and regular density matrix types in python will prove to be very valuable further down the road. In this PR I will tackle this functionality.

**Related issues**
- closes #677 

**Additional context**
I have removed the `spinDensity` binding from `linearExpansion`, since the `SpinResolvedOneDM` binding can now calculate this for us. I have, however, added the `calculateSpinResolved1DM` and `calculateSpinResolved2DM` functionalities to the `linearExpansion` bindings. This grants a lot of freedom to manipulate the density matrices using the python bindings. 

**Don't forget:**

### DIFF
--- a/gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp
+++ b/gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp
@@ -79,7 +79,7 @@ void bindUHFSCFEnvironment(py::module& module) {
             })
 
 
-        // Define read-only 'getters'.
+        // Define read-only 'getters'
         .def_readonly(
             "coefficient_matrices_alpha",
             &GQCP::UHFSCFEnvironment<double>::coefficient_matrices_alpha)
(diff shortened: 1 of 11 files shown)


xdvriend:
@lelemmen , with these python bindings it is possible to return `spinResolvedDM's` in python. Should I rework the UHF bindings again in order that the `def_readonly()` bindings return the spin resolved DM and keep the alpha and beta DM bindings as separate functions? Or don't you think this is necessary?

xdvriend:
The UHF `def_readonly()`getters can be added for the Fock matrices as well, but for that `USQOneElectronOperator` needs python bindings. Since we are in the midst of a large operator refactor, I will leave this as it is for now.

gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp
```diff
@@ -88,6 +88,19 @@ void bindUHFSCFEnvironment(py::module& module) {
             "coefficient_matrices_beta",
             &GQCP::UHFSCFEnvironment<double>::coefficient_matrices_beta)
 
+        .def_readonly(
+            "spin_resolved_density_matrices",
```
lelemmen, line 92: Is there possibly a reason why the Python method isn't called `density_matrices`?

lelemmen (approved):
Great addition to the code, thanks!

gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp
xdvriend: I thought this would convey more clearly what the method returned, since we also have bindings for the separate alpha and beta matrices. But I can rename it to just `density_matrices` if you'd prefer that.

Issue #688, opened 2020-09-23
Compile-time correctness


lelemmen:
**Introduction**
One of the defining features of this C++ library is compile-time correctness. In easy words: _what is implemented in the C++ library should be relatable to a specific part of electronic structure theory_. Since we can never aim for "having implemented everything", we cannot use run-time polymorphism using pure virtual methods and overrides, because if only one implementation were missing (like some obscure matrix-vector product that is only used for one of the deriving types), we'd have to throw a runtime error, possibly wasting hours of compute.

This issue will track the progress on implementing this behavior, since I'm not comfortable shipping a v1.0.0 without it.

This is a multi-PR, multi-month (and hopefully, multi-contributor) project. Once it is achieved, we'll be further increasing the gap between other software packages and ours, since the connection between the theory and the code library is further deepened. It won't be long until we have implemented electronic structure's domain specific language.

**When is this issue finished?**
- [x] When all types have been scrutinized for their R/U/G behavior;
- [x] When the CI module implementation no longer uses pure virtual methods;
- [x] When the Python bindings have been adjusted to accommodate all the changes;
~- [ ] When the website's documentation covers all the changes that are related to this issue.~ This is transferred to #607.

lelemmen:
Part 1: Implementing the back bone `SimpleSQOneElectronOperator` for R/G one-electron operators.
--- 

This PR (#686) will gather work on refactoring the code such that it is more compile-time correct. We had been using `SQOneElectronOperator` and `OneDM` to refer to either the restricted or generalized form, but this leads to potentially mixing restricted one-electron operators with generalized density matrices, something our code isn't prepared for. (Note that the theory doesn't disallow this mixing, it just requires a different implementation.)

lelemmen:
Part 2: BasisTransformable and GSQOneElectronOperator
---

In this PR, I've introduced the `BasisTransformable` interface. Unfortunately, `SimpleSQOneElectronOperator` now implements two interfaces, which both inherit from `CRTP`, which causes a diamond pattern and therefore an ambiguous `derived()` method. In order to resolve this issue, I'll opt for removing the `CRTP` as a base class for the interfaces, having to use the appropriate `static_cast<const&>` or `static_cast<&>` wherever needed.

lelemmen:
Part 3: Density matrices
---

There are a couple of things I don't really like about the current density matrices, which are the subject of this PR:
1. Their names start with 'One' and 'Two' instead of '1' and '2'.
[...]

PR #694, opened 2020-09-25
Further operator refactors.

lelemmen:
**Short description**
This is part 2 of the compile-time correctness refactor #688.

**Related issues**

**Additional context**

**Don't forget:**


xdvriend (approved):
I really like the transformation interface! This will prevent so much code duplication.

PR #697, opened 2020-09-29
Density & transformation matrix improvements

lelemmen:
**Short description**
This is part 3 of #688.

**Related issues**

**Additional context**

**Don't forget:**
- [x] if new files are created: update the Doxygen file, the collective gqcp.hpp header and the CMake files

PR #698, opened 2020-09-30
Transformation matrices

### LINKED ISSUE #413: Better interfaces for SquareMatrix
`SquareMatrix` is a child class of `Matrix`, meant to enable operations only suitable for square matrices.

Currently, we can construct a zero- or ones-initialized square matrix as 

```cpp
SquareMatrix<double> M = SquareMatrix<double>::Zero(2, 2);
SquareMatrix<double> I = SquareMatrix<double>::Identity(2, 2);
```

- [x] I think it would make more sense to disable these named constructors and opt for the following ones instead:

```cpp
SquareMatrix<double> M = SquareMatrix<double>::Zero(2);
SquareMatrix<double> I = SquareMatrix<double>::Identity(2);
```

- [x] A constructor that creates a `RandomSymmetric` `SquareMatrix` would also be a nice addition.

lelemmen:
**Short description**
This is part 4 and part 5 of #688.

**Related issues**
This PR closes #413.

**Additional context**

**Don't forget:**


lelemmen:
With the latest commit, the failing tests introduced in #697 should be fixed.

guacke (approved):

Issue #689, opened 2020-09-24
Improved Python bindings for templated classes


lelemmen:
**Describe the feature you'd like**
In native Python, we're never worried about exposing an underlying type to a class. For example, `np.array` is suited to work with any underlying data type (floats, doubles, complex numbers, ...). I'd like this library's Python bindings to follow the same approach.

**Describe what the current code offers in relation to what it lacks**
Currently, we have to expose every template class explicitly, using a suffix for the Python class name. This leads to potentially very long names, like `LinearExpansion_SpinUnresolved`, which isn't very Pythonic nor clean in my opinion.

**Additional context**
I'm sure it can be done, but I don't know how, yet, so I think it's good to start with a small prototype.

PR #699, opened 2020-09-30
Complex HF bindings

### LINKED ISSUE #645: Complex GHF python bindings
For GHF, using complex values is sometimes a necessity. It would be useful to include this feature in the python bindings.

The current GHF python bindings only support real valued GHF.

xdvriend:
In this PR I will implement the ability to use the python bindings to perform complex Hartree-Fock calculations.

**Short description**
By adjusting the python bindings in such a way that they can accept `complex double`s, complete HF calculations should become possible.

**Related issues**
- closes #645 

**Additional context**
I will add this ability to all HF methods:

- [x] GHF
- [ ] RHF
- [ ] UHF

I will also add an example notebook showing how to use the complex HF python bindings.

- [ ] Example notebook

I will look into doing this in a way that accommodates issue #689, however I'm not sure at the moment how to approach this.

**Don't forget:**


xdvriend:
At the moment, complex bindings don't work yet. The problem lies within the constructor for the environment not accepting a complex `Eigen::MatrixXcf`as an argument. I'll look further into this.

xdvriend:
Okay, I have run into an issue that makes implementing the complex bindings very tricky.

Since the bindings use a template argument `<Scalar>`, all arguments of the constructors etc. will also have this argument, which either becomes `double` or `complex<double>`. 

The problem now lies with the `SQHamiltonian`. I can't make the Hamiltonian complex, but I can't set it on `double` because this causes a problem when trying to create a `GHFSCFEnvironment<complex<double>>`. I have tried adjusting the `.hpp`files of the Hamiltonian and the GHF methods to make this possible, but so far without any success. 

Any advice on how I can solve this is more than welcome (@guacke, @lelemmen), and I'll keep looking into this as well, whenever I have some time.

xdvriend:
This PR is hopelessly outdated, so I'm closing it.

PR #700, opened 2020-09-30
ONVPath evaluation one electron integrals.

dariatols:
**Short description**
Evaluation of the one-electron integrals using `ONVPath`.

**Related issues**
The open (and possibly closed) issues that are related to this PR.

**Don't forget:**


lelemmen:
@dariatols Could you remove the extra `yarn.lock` file that you pushed?

dariatols:
@lelemmen This method now produces exactly the same results as the "old" `evaluate(const ScalarSQOneElectronOperator<double>& one_op, const bool diagonal_values) const` method.

dariatols:
I could write an additional `evaluate` method where an instance of `MatrixRepresentationEvaluationContainer<_Matrix>&` is given as argument too, but I don't see the possible added value.

xdvriend:
I have written the python bindings using a templating approach in order to allow for spin resolved and spin unresolved ONVBases to be used. At the moment, the `SpinResolved`bindings are commented out, since the constructor won't work, as `SpinResolvedONVBasis`does not have a member `ONV`.

xdvriend (approved):
At first glance, the new evaluate method looks a lot more readable and clear than the previous one! I'll take some time to make sure I completely understand the way it works. But for now, I'll already approve this PR.

lelemmen:
> I could write an additional `evaluate` method where an instance of `MatrixRepresentationEvaluationContainer<_Matrix>&` is given as argument too (as is the case with the "old" `evaluate`method of Tobias), but I don't see the possible added value.

That's for the high-level (API) CI refactor. We'll tackle that later.

gqcp/include/ONVBasis/SpinUnresolvedONVBasis.hpp
```diff
@@ -311,6 +315,69 @@ class SpinUnresolvedONVBasis:
...
+                    size_t address = onv_path.addressAfterCreation(onv_path.orbitalIndex(), onv_path.electronIndex());
+
+                    double h_pq = onv_path.sign() * one_op_par(onv_path.orbitalIndex(), q);
+
+                    // Add the one-electron integral as matrix elements of a Hermitian matrix.
+                    ONV_iterator.addColumnwise(address, h_pq);
+                    ONV_iterator.addRowwise(address, h_pq);
+
+                    // Move orbital index such that other unoccupied orbitals can be found within the loop.
+                    onv_path.addVertical();
```
lelemmen, line 353: I'm not sure I like this call at the end of the loop, as it creates a path with an invalid state. We'll discuss this later.

lelemmen (approved):
GREAT job on this PR!!! 🎉 

Could you check if the documentation needs a small update due to the name changes you've done? Afterwards, you're free to merge this one!

---
I'm hereby awarding you the following badge. Wear it with pride! 

C++-first-PR-Daria

PR #764, opened 2020-11-02
PR: Change class to struct for traits.

### LINKED ISSUE #706: Use struct for traits
**Describe the current state of a part of the code and a change that would improve the code base**
In the current code, a trait is alternatively defined as a `struct` or `class`. In my opinion, it would be better to implement them all as `struct`s as their members should be stored as `public`.

dariatols:
**Short description**
In the current code, a trait is alternatively defined as a `struct` or `class`. In my opinion, it would be better to implement them all as `struct`s as their members should be stored as `public`.

**Related issues**
Closes #706.

**Don't forget:**


lelemmen (approved):
Yup, that's good!

PR #766, opened 2020-11-03
Clean up old source files. Closes #761.

### LINKED ISSUE #761: Remove unused source code files
After #707, some of the library's functionality was disabled by commenting out CMake include statements. Those files that are just available in the source code, but aren't used, should be transferred to the corresponding issues on GitHub in order to track this technical debt.

lelemmen:
This PR cleans up previously commented-out source code. I've added the corresponding parts to their respective issues.

**Don't forget:**


lelemmen:
@xdvriend This is just a pretty small PR that only deletes a couple of old files that were still in the library.

xdvriend (approved):
Allright, looks good to me!

PR #768, opened 2020-11-04
Make Eigenpair a template class.

### LINKED ISSUE #538: Use a template argument in Eigenpair for the scalar representation
Currently, `Eigenpair` is a class that can only work with real numbers.

The code would benefit from making `Eigenpair` a class template, using a template parameter that represents the scalar representation of the elements of the eigenvector.

**Additional context**
We should also make a decision if also want to add a template parameter for the scalar representation of the eigenvalue.

dariatols:
**Related issues**
Closes #538.

**Don't forget:**


gqcp/include/Mathematical/Optimization/Eigenproblem/Eigenpair.hpp
```diff
@@ -19,56 +19,84 @@
...
+        Eigenpair(0.0, VectorX<double>::Zero(dimension)) {}
 
 
     // PUBLIC METHODS
 
     /**
+     * A function that returns the eigenvalue associated to this eigenpair.
+     * 
      *  @return the eigenvalue associated to this eigenpair
      */
```
lelemmen, line 67:
```suggestion
    /**
     *  @return The eigenvalue associated to this eigenpair.
     */
```

I know that I mentioned to add extra explanation to member functions, but for trivial accessors, I don't think is required.

Also, "A function that ... " seems pretty redundant as documentation style.

gqcp/include/Mathematical/Optimization/Eigenproblem/Eigenpair.hpp
```diff
@@ -19,56 +19,84 @@
...
+     * 
      *  @return the eigenvector associated to this eigenpair
      */
-    const VectorX<double>& eigenvector() const { return this->m_eigenvector; };
+    const VectorX<Scalar>& eigenvector() const { return this->m_eigenvector; };
 
     /**
-     *  @param other            the other Eigenpair
-     *  @param tolerance        a tolerance for comparison
+     * Checks if this Eigenpair is equal to the other Eigenpair.
```
lelemmen, line 78:
```suggestion
     *  Check if this Eigenpair is equal to the other Eigenpair.
```

We adopt an imperative documentation style.

gqcp/src/Mathematical/Optimization/CMakeLists.txt
```diff
@@ -1,3 +1,2 @@
-add_subdirectory(Eigenproblem)
```
lelemmen, line 1: Good catch. Could you also remove the related directory from `cmake/Doxyfile.in`?

lelemmen (approved):
Looks good to me!

Feel free to merge after updating the Doxyfile.

PR #782, opened 2020-11-10
Improve arithmetic operators on SQHamiltonian.

### LINKED ISSUE #745: Implement addition/multiplication operators on `SQHamiltonian`
Currently, we can only add restricted operators to restricted Hamiltonians. We'd like to improve this behaviour by implementing support for adding restricted/generalized operators to their Hamiltonian counterparts. (This can be done using the templates, so this isn't too much work.)

lelemmen:
Closes #745.

**Don't forget:**


xdvriend (approved):
👍

PR #705, opened 2020-10-21
Calculate 1DM with ONVPath API.

dariatols:
**Short description**
The calculation of the one-electron density matrix will be refactored such that it will make use of the `ONVPath` API.

**Related issues**
https://github.com/GQCG/GQCP/issues/687 https://github.com/GQCG/GQCP/issues/648
closes https://github.com/GQCG/GQCP/issues/719

**Don't forget:**
- [ ] if new files are created: update the Doxygen file, the collective gqcp.hpp header and the CMake files

PR #783, opened 2020-11-10
Implement 1-DM for linear expansions in spin-unresolved ONV bases

### LINKED ISSUE #719: Implement 1-DM for spin-unresolved linear expansions
We would like to be able to implement the 1-DM for spin-unresolved linear expansions, i.e. for `LinearExpansion<SpinUnresolvedONVBasis>`.

lelemmen:
- [ ] closes #719;

@dariatols The previous implementation from #705 was very wrong. You were emplacing elements in an ONV-basis-dimensional object. I've tried to improve upon the algorithm, but without any success yet. The reason your tests weren't picking up any errors was just plain-old luck in my opinion. Feel free to adjust any estimates according to your thoughts.

**Don't forget:**

### DIFF
--- a/gqcp/include/QCModel/CI/LinearExpansion.hpp
+++ b/gqcp/include/QCModel/CI/LinearExpansion.hpp
@@ -46,35 +46,33 @@ namespace GQCP {
 /**
  *  A class that represents a linear expansion inside an ONV basis.
  * 
- *  @tparam _ONVBasis           the type of ONV basis
+ *  @tparam _ONVBasis           The type of ONV basis.
  */
 template <typename _ONVBasis>
 class LinearExpansion {
-
 public:
+    // The type of the ONV basis.
     using ONVBasis = _ONVBasis;
 
 
 private:
-    ONVBasis onv_basis;              // the ONV basis with respect to which the coefficients are defined
-    VectorX<double> m_coefficients;  // the expansion coefficients
+    // The ONV basis with respect to which the coefficients are defined
+    ONVBasis onv_basis;
+
+    // The expansion coefficients.
+    VectorX<double> m_coefficients;
 
 
 public:
     /*
-     *  CONSTRUCTORS
+     *  MARK: Constructors
      */
-
-    /*
-     *  The default constructor.
-     */
-    LinearExpansion() = default;
 
     /**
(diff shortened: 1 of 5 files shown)


dariatols:
@lelemmen The current code will result in an error since I have not yet disabled the tests where `SpinUnresolved`'s `calculate1DM()` is compared to `calculateNDMElement`. More specifically, 
```
error: in "calculate1DM_SpinUnresolved_NDM": check std::abs(linear_expansion.calculateNDMElement({4}, {4}) - D_specialized(4, 4)) < 1.0e-12 has failed
```
will occur. Something seems to be going wrong with the `sign` of the given `ONV`s in this method.

dariatols:
On the bright side: `SpinUnresolved`'s `calculate1DM()` now produces the same 1DM as a `SpinResolved` `ONVBasis` with 0 beta electrons 🎉.

dariatols:
To do:
- [ ] I still don't understand why `onv_path.sign()` will work fine for determining the sign while it does not take the sign of the creation operator into account. I have written a new function `signAfterCreation(const size_t p)` in `ONVPath` but this one will give the "wrong" sign..

lelemmen:
@dariatols I don't know, you tell me!

dariatols:
IMG_0040

dariatols:
@lelemmen I was too focused on the result of an annihilation/creation on an ONV. These operators do not have the same effect on ONVs and ONVPaths. See note above for an in-depth explanation with ONVPaths.

dariatols:
And I cannot choose you as reviewer since you've made this PR.

gqcp/include/QCModel/CI/LinearExpansion.hpp
```diff
@@ -541,49 +572,79 @@ class LinearExpansion {
...
+                    const auto p = onv_path.orbitalIndex();
+
+                    const double value = c_I * c_J;
+
+                    D(p, q) += onv_path.sign() * value;
+                    D(q, p) += onv_path.sign() * value;
+
+                    // Move orbital index such that other unoccupied orbitals can be found within the loop.
+                    onv_path.leftTranslateVerticalArc();
+                }
```
lelemmen, line 635: @dariatols This code is so clean, nice!

PR #812, opened 2020-11-26
Adding functionality for the constrained CI research

xdvriend:
**Short description**
In this PR I'll add the missing functionality necessary to perform the constrained CI calculations.

**Additional context**
At the moment I've encountered some problems with this.

- ~~When I create a `def_property`binding, The components are no longer callable. E.g.: `u_mulliken_operator.alpha()` gives an error.~~
- ~~The `*=` operator still gives the error: `can't assign to function call`.~~

**Don't forget:**


xdvriend:
This small PR fixes the problems I ran into while trying to run the spin constrained calculations for the constrained CI research. Everything should be as discussed today.

lelemmen:
@xdvriend Could you set up this PR for Q/A? I think the spin-redistributions epic (#663) is finished, so I'd like to close it!

xdvriend:
That's strange, it was already in the Q/A pipeline. I think it was moved when this PR was added to the new sprint. I'll put it back!

lelemmen (approved):

Issue #827, opened 2020-12-01
Python bindings for inverse and transformed SimpleTransformation


dariatols:
**Describe the feature you'd like**
In my Fukui-Dyson research, two transformation matrices must be transformed into a compound transformation matrix. The consequence of this is the need for additional Python bindings for `SimpleTransformation<...>::inverse()` and `SimpleTransformation<...>::transformed()`.

**Describe what the current code offers in relation to what it lacks**
The current code only has a binding for `SimpleTransformation::matrix()` and should be complemented by ìnverse()` and `transformed()`.

**Additional context**
See https://github.com/GQCG-res/fukui_dyson_outline/issues/34.

PR #828, opened 2020-12-01
Additional bindings regarding Transformations.

dariatols:
**Short description**
This PR introduces bindings for 
- [x] `inverse()`
- [x] `transformed()`

in `SimpleTransformation`.

**Related issues**
Closes https://github.com/GQCG/GQCP/issues/827.

**Don't forget:**


gqcpy/include/interfaces.hpp
```diff
@@ -472,7 +472,19 @@ void bindSimpleTransformationInterface(Class& py_class) {
...
+             &Type::inverse,
+             "Return the inverse transformation of this transformation matrix.")
+
+        .def(
+            "transformed",
+            [](const Type& T_this, const Type& T) {
+                return T_this.transformed(T);
+            },
+            py::arg("T"),
+            "The transformation that encapsulates the sequential application of this transformation, followed by the given transformation.");
```
lelemmen, line 487: While this a good addition to the code, could you perhaps replace it with `bindBasisTransformableInterface`?

dariatols: That is a much cleaner solution indeed, and it prevents duplication of code.

dariatols:
@lelemmen Is it possible that `transformed` and  `rotated` were wrongly binded as these were returning `None` objects instead of a `Transformation` matrix?

gqcpy/include/interfaces.hpp
```diff
@@ -472,7 +472,21 @@ void bindSimpleTransformationInterface(Class& py_class) {
...
              &Type::matrix,
-             "Return the transformation matrix that collects the expansion coefficients of the new basis (vectors) in the old basis as columns.");
+             "Return the transformation matrix that collects the expansion coefficients of the new basis (vectors) in the old basis as columns.")
+
+        .def("inverse",
+             &Type::inverse,
+             "Return the inverse transformation of this transformation matrix.")
+
+        .def(
+            "isUnitary",
```
dariatols, line 482: I've added this binding since I might possible be using this for telling my story in a notebook.

gqcpy/include/interfaces.hpp
```diff
@@ -472,7 +472,21 @@ void bindSimpleTransformationInterface(Class& py_class) {
...
+            [](const Type& T, const double threshold = 1.0e-12) {
+                return T.isUnitary(threshold);
+            },
```
lelemmen, line 485:
```suggestion
            &Type::isUnitary,
            py::arg("threshold") = 1.0e-12,
```

I think this can be written shorter.

gqcpy/src/Basis/Transformations/UTransformation_bindings.cpp
```diff
@@ -41,9 +41,23 @@ void bindUTransformation(py::module& module) {
...
+        .def("inverse",
+             &UTransformation<double>::inverse,
+             "Return the inverse transformation of this transformation matrix.")
+
+        .def(
+            "isUnitary",
+            [](const UTransformation<double>& T, const double threshold = 1.0e-12) {
+                return T.isUnitary(threshold);
+            },
+            "Return if this transformation is unitary, within the given threshold");
```
lelemmen, line 54: This is a copy-paste from the implementation in `bindBasisTransformableInterface`. Could you wrap these methods into a re-usable function and place it into `interfaces.hpp`?

lelemmen (commented):
Good additions to the bindings, but I think you could make them more future-proof if you implemented a re-usable interface binding method.

gqcpy/src/Basis/Transformations/UTransformation_bindings.cpp
dariatols: It was a copy-paste from `SimpleTransformationInterface`, but since `UTransformation` is a `SpinResolved`-like object, I could not use this interface. I've made a new interface consisting of bindings for SpinResolved-like transformations like `UTransformation` is.

lelemmen: I think I didn't express myself clearly enough.

In your current implementation, there are still 2 occurrences of a `"isUnitary"` and an `"isInverse"` binding. Could you create a function, for example `bindBasisTransformableOperationsInterface`, collecting these two bindings? You can then use this function inside the implementation of `bindSimpleTransformationInterface` and your new binding.

If this doesn't make sense, let me know and we'll arrange a short meeting to discuss this.

dariatols: This is indeed a better way, and is now incorporated in the last commit.

lelemmen (approved):
This is super clean! Nice work.

Issue #811, opened 2020-11-26
Improve the current Python bindings by refactoring common interfaces


lelemmen:
There are probably a couple of cases where we have just copy-pasted Python bindings. We should make sure that this copy-pasting is avoided, by using reusable interfaces.

Issue #794, opened 2020-11-19
Stability analysis python bindings


xdvriend:
**Describe the feature you'd like**
Once all features of the stability analysis have been implemented, it would be beneficial to create python bindings for these API's as well. This could go hand in hand with a example notebook of the `Cascading Hartree-Fock algorithm` which could be a possible subject for a blog post on the website as well.

PR #834, opened 2020-12-03
Refactor HF python bindings

### LINKED ISSUE #829: Python bindings for stability checks
The stability analysis is now implemented in GQCP. The next step would be to be able to perform these calculations using the python bindings as well.

**Additional context**
This is part 1 of the stability analysis python bindings issue #794 . Once issue #793 is completed, #794 can be finished completely.

This part of issue #794 wasn't really blocked by the ability to follow internal instabilities. Hence why I split up #794 into two separate issues and changed the original issue to a small epic.

xdvriend:
**Short description**
This PR aims to refactor the python bindings of the HF `QCModel` while avoiding code duplication. I will also add the python bindings for the stability checks.

**Related issues**
- This is related to #811 and #794
- This closes #829

**Additional context**

- [x] Create interface for `QCModel::R/U/GHF`.

- [x] Create Interface for stability analysis.

- [x] Create R/U/GHFStabilityMatrices bindings.

- [x] Add stability analysis example notebook.

**Don't forget:**


xdvriend:
At the moment, everything works in python except a few methods of `QCModel::UHF`.

Some API's of `QCModel::UHF` return a `SpinResolved` type, which is not recognised by python. I can't let it inherit from the 'SpinResolvedInterface' as UHF itself is not a `SpinResolved` class. I also don't like the idea of implementing python classes foor 'SpinResolvedVector' or anything like that. I'm looking into a solution.

xdvriend:
Other than the main goal of this PR, some smaller changes were made as well. Here's an explanation as to why I made those changes:

- I added a public alias `Hamiltonian` to the `QCModels` in order to add `calculateStabilityMatrices` to the interface.
- I changed `2 * g(a,i,j,b)` to `g(a,i,j,b) + g(ani,j,b)` in the RHF stability matrix calculation because the former wouldn't work for the complex case.
- I changed `.dimension(0)` to `.rows()` in. the construction of the block represented stability matrices, as `.dimension(0)` did not work for the complex case.
- I changed the `.external()` method in `GHFStabilityMatrices` to `.realComplex()` in order to be conform to the other two methods.
- I made it so some functions in the `QCModelUHF` return a `SpinResolved` type, to be more conform with the other two.
- `.transformed()` and `.rotated()` now return the transformed type instead of a none type.
- I added an interface for the `QCModel` python bindings and two for the stability calculations, as it was the cleanest option to split the interface for real and complex.
- The addition of the spin resolved bindings is explained in my previous comment.

The build still fails because of the `Simple 2DM test`. That test works fine locally so I've got no idea why it doesn't work here. I'm looking into it.

gqcp/include/QCModel/HF/RHF.hpp
```diff
@@ -447,7 +451,7 @@ class RHF {
             for (const auto& a : orbital_space.indices(OccupationType::k_virtual)) {
                 for (const auto& j : orbital_space.indices(OccupationType::k_occupied)) {
                     for (const auto& b : orbital_space.indices(OccupationType::k_virtual)) {
-                        singlet_A_slice(i, a, j, b) = 2 * g(a, i, j, b) - g(a, b, j, i);
+                        singlet_A_slice(i, a, j, b) = g(a, i, j, b) + g(a, i, j, b) - g(a, b, j, i);
```
lelemmen, line 454: I fail to see why the new implementation produces correct results, and the old one doesn't. Both formulas are exactly the same.

gqcp/include/QCModel/HF/UHF.hpp
```diff
@@ -409,8 +413,8 @@ class UHF {
 
         const auto K_a = this->numberOfSpinOrbitals(Spin::alpha);
         const auto K_b = this->numberOfSpinOrbitals(Spin::beta);
-        const auto N_a = this->numberOfElectrons(Spin::alpha);
-        const auto N_b = this->numberOfElectrons(Spin::beta);
+        const auto N_a = this->numberOfElectrons().alpha();
+        const auto N_b = this->numberOfElectrons().beta();
```
lelemmen, line 417: Since you've made this change, could you change the `private` numbers of electrons to a `SpinResolved` type as well?

gqcpy/include/interfaces.hpp
```diff
@@ -89,7 +90,7 @@ void bindBasisTransformableInterface(Class& py_class) {
         .def(
             "transformed",
             [](Type& transformable, const Transformation& T) {
```
lelemmen, line 92: I noticed that the first argument may become `const`.

```suggestion
            [](const Type& transformable, const Transformation& T) {
```

gqcpy/include/interfaces.hpp
```diff
@@ -699,4 +700,145 @@ void bindSimple2DMInterface(Class& py_class) {
...
+    using Type = typename Class::type;
+    using Hamiltonian = typename Type::Hamiltonian;
+    using Scalar = typename Type::Scalar;
+
+    py_class
+        .def(
+            "calculateOrthonormalBasis1DM",
+            [](const Type& parameters) {
+                return parameters.calculateOrthonormalBasis1DM();
+            },
```
lelemmen, line 727:
```suggestion
            &Type::calculateOrthonormalBasis1DM,
```

I think this implementation can become shorter.

gqcpy/include/interfaces.hpp
```diff
@@ -699,4 +700,145 @@ void bindSimple2DMInterface(Class& py_class) {
...
+            "Return the 1-DM expressed in an orthonormal basis related to these optimal HF parameters.")
+
+        .def(
+            "calculateScalarBasis1DM",
+            [](const Type& parameters) {
+                return parameters.calculateScalarBasis1DM();
+            },
```
lelemmen, line 734:
```suggestion
            &Type::calculateScalarBasis1DM,
```

gqcpy/include/interfaces.hpp
```diff
@@ -699,4 +700,145 @@ void bindSimple2DMInterface(Class& py_class) {
...
+            "Return the HF 1-DM in the scalar/AO basis related to these optimal parameters.")
+
+        .def(
+            "calculateStabilityMatrices",
+            [](const Type& parameters, const Hamiltonian& hamiltonian) {
+                return parameters.calculateStabilityMatrices(hamiltonian);
+            },
```
lelemmen, line 741:
```suggestion
            &Type::calculateStabilityMatrices,
            py::arg("hamiltonian"),
```

I've added a `py::arg` for the Hamiltonian argument. Also, I think (if this solution works), you can remove the `using Hamiltonian = ...` aliases in the HF QCModels.

gqcpy/include/interfaces.hpp
```diff
@@ -699,4 +700,145 @@ void bindSimple2DMInterface(Class& py_class) {
...
+
+        .def("expansion",
+             &Type::expansion,
+             "Return the coefficient matrix that expresses every spatial orbital (as a column) in its underlying scalar basis.")
+
+        .def(
+            "numberOfElectrons",
+            [](const Type& parameters) {
+                return parameters.numberOfElectrons();
+            },
```
lelemmen, line 752:
```suggestion
            &Type::numberOfElectrons,
```

gqcpy/include/interfaces.hpp
```diff
@@ -699,4 +700,145 @@ void bindSimple2DMInterface(Class& py_class) {
...
+    // The C++ type corresponding to the Python class.
+    using Type = typename Class::type;
+    using Scalar = typename Type::Scalar;
+
+    py_class
+        .def(
+            "internal",
+            [](const Type& stability_matrices) {
+                return stability_matrices.internal();
+            },
```
lelemmen, line 788:
```suggestion
            &Type::internal,
```

gqcpy/include/interfaces.hpp
```diff
@@ -699,4 +700,145 @@ void bindSimple2DMInterface(Class& py_class) {
...
+            "Return the internal HF stability matrix.")
+
+        .def(
+            "isInternallyStable",
+            [](const Type& stability_matrices, const double threshold) {
+                return stability_matrices.isInternallyStable(threshold);
+            },
```
lelemmen, line 795:
```suggestion
            &Type::isInternallyStable,
```

gqcpy/include/interfaces.hpp
```diff
@@ -699,4 +700,145 @@ void bindSimple2DMInterface(Class& py_class) {
...
+            py::arg("threshold") = 1.0e-05,
+            "Return a boolean, telling us if the real or complex valued internal stability matrix belongs to a stable or unstable set of parameters.")
+
+        .def(
+            "printStabilityDescription",
+            [](const Type& stability_matrices) {
+                py::scoped_ostream_redirect stream(std::cout, py::module::import("sys").attr("stdout"));
```
lelemmen, line 802:
```suggestion
                py::scoped_ostream_redirect stream {std::cout, py::module::import("sys").attr("stdout")};
```

gqcpy/include/interfaces.hpp
```diff
@@ -699,4 +700,145 @@ void bindSimple2DMInterface(Class& py_class) {
...
+    // The C++ type corresponding to the Python class.
+    using Type = typename Class::type;
+    using Scalar = typename Type::Scalar;
+
+    py_class
+        .def(
+            "realComplex",
+            [](const Type& stability_matrices) {
+                return stability_matrices.realComplex();
+            },
```
lelemmen, line 828:
```suggestion
            &Type::realComplex,
```

gqcpy/include/interfaces.hpp
```diff
@@ -699,4 +700,145 @@ void bindSimple2DMInterface(Class& py_class) {
...
+            "Return the real->complex external HF stability matrix.")
+
+        .def(
+            "isExternallyStable",
+            [](const Type& stability_matrices, const double threshold) {
+                return stability_matrices.isExternallyStable(threshold);
+            },
```
lelemmen, line 835:
```suggestion
            &Type::isExternallyStable,
```

gqcpy/src/QCModel/HF/RHF_bindings.cpp
```diff
@@ -28,43 +30,36 @@ namespace gqcpy {
...
         .def(
             "spinOrbitalEnergiesInterleaved",
-            &QCModel::RHF<double>::spinOrbitalEnergiesInterleaved,
+            &QCModel::RHF<Scalar>::spinOrbitalEnergiesInterleaved,
             "Return all the spin-orbital energies, with the alpha spin-orbital energies appearing before the beta spin-orbital energies.");
+
+    // Expose the `HartreeFock` interface.
+    bindQCModelHartreeFockInterface(py_QCModelRHF);
+}
+
```
lelemmen, line 55:
```suggestion


```

gqcpy/src/QCModel/HF/StabilityMatrices/GHFStabilityMatrices_bindings.cpp
```diff
@@ -0,0 +1,83 @@
...
+    bindQCModelGHFStabilityInterface(py_GHFStabilityMatrices_d);
+
+    // Define Python classes related to `complex GHFStabilityMatrices` and expose their interfaces.
+    py::class_<GHFStabilityMatrices<complex>> py_GHFStabilityMatrices_cd {module, "GHFStabilityMatrices_cd", "The complex generalized Hartree-Fock stability matrices."};
+
+    // Expose the `HartreeFockComplexStability` interface.
+    bindQCModelHartreeFockComplexStabilityInterface(py_GHFStabilityMatrices_cd);
+
+    // Expose the `GHFStability` interface.
+    bindQCModelGHFStabilityInterface(py_GHFStabilityMatrices_cd);
```
lelemmen, line 79: I really like this! For the complex types, the interface has to be different, which is clearly indicated by the implementation of these bindings.

gqcpy/src/QuantumChemical/Spin_resolved_bindings.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+
+// Provide some shortcuts for frequent namespaces.
+namespace py = pybind11;
+using namespace GQCP;
+
+template <typename Of>
+void bindSpinResolved(py::module& module, const std::string& name, const std::string& description) {
+
+    // Define the Python class for `spinResolved` types.
```
lelemmen, line 34:
```suggestion
    // Define the Python class for `SpinResolved` types.
```

gqcpy/src/QuantumChemical/Spin_resolved_bindings.cpp
```diff
@@ -0,0 +1,52 @@
...
+    py::class_<SpinResolved<Of>> py_spinResolved {module, name.c_str(), description.c_str()};
+
+    // Expose the `BasisTransformable` API to the Python class.
```
lelemmen, line 37:
```suggestion
    // Expose the `SpinResolvedBase` API to the Python class.
```

gqcpy/src/QuantumChemical/Spin_resolved_bindings.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+
+/**
+ *  Bind all types of `SpinResolved`s.
+ */
+void bindSpinResolvedTypes(py::module& module) {
+
+    bindSpinResolved<size_t>(module, "SpinResolvedLong", "A spin resolved encapsulation of two unsigned longs.");
+    bindSpinResolved<std::vector<double>>(module, "SpinResolvedStdVector", "A spin resolved encapsulation of two std::vectors.");
+    bindSpinResolved<VectorX<double>>(module, "SpinResolvedVector", "A spin resolved encapsulation of two GQCP::Vectors.");
```
lelemmen, line 49:
```suggestion
    bindSpinResolved<size_t>(module, "SpinResolved_size_t", "A spin resolved encapsulation of two unsigned longs.");
    bindSpinResolved<std::vector<double>>(module, "SpinResolved_std_vector_d", "A spin resolved encapsulation of two std::vectors.");
    bindSpinResolved<VectorX<double>>(module, "SpinResolved_VectorX_d", "A spin resolved encapsulation of two GQCP::Vectors.");
```

I'd like to use the convention that the template arguments are concatenated using `_`. Of course, this horrible naming has its origin related to #689.

gqcpy/src/QuantumChemical/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 list(APPEND python_bindings_sources
     ${CMAKE_CURRENT_SOURCE_DIR}/Spin_bindings.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/Spin_resolved_bindings.cpp
```
lelemmen, line 3: Could you maybe rename the file to `SpinResolved_bindings`?

lelemmen (commented):
Great job on rewriting the HF Python bindings! Improving on the re-usable interfaces is really appreciated, since it makes the actual Python bindings easier to read.

There's only a little more work to be done before merging this PR.

gqcp/include/QCModel/HF/RHF.hpp
xdvriend: This was done because the old implementation resulted in an error when trying to bind the complex stability methods. The operation `*` was not compatible between a `complex double` and an `integer`. I found this really strange myself, but since it was an issue of type compatibility, this fixed it. 

I'm not a big fan of it either, so if you have any other suggestions to fix this, I'd be happy to change it. But as long as the complex bindings are enabled, `2 * g(a, i, j, b)` won't compile.

gqcpy/include/interfaces.hpp
xdvriend: Writing it this way will give an error as pybind won't be able to recognise which of the overloaded `numberOfElectrons()` API's of the RHF model it is supposed to use.

gqcpy/include/interfaces.hpp
xdvriend: I initially did all of the bindings this way, but some gave the error where pybind wasn't able to figure out which API it was supposed to bind. I'll look into it again and use the shorter notation where possible. I will of course do this for all `StabilityMatrices` bindings and not just for this particular one.

gqcp/include/QCModel/HF/RHF.hpp
xdvriend: Screenshot 2020-12-07 at 13 41 25

This is a screenshot of the error when `2 * g(a, i, j, b)` is used.

lelemmen: Oh, I see. I think you can use `2.0` (as a literal `double`) and then it will compile.

On another note. Have you double-checked the correctness of these formulas in the complex case? (From experience, I know that a prefactor `2` in the real case often results from the permutational symmetry of **real** integrals.)

gqcpy/include/interfaces.hpp
xdvriend: If written this way, I get the following error:

Screenshot 2020-12-07 at 13 56 25

I believe this may be the case because the `QCModel` contains a `public` and a `static` `calculateOrthonormalBasis1DM()` API.

gqcpy/include/interfaces.hpp
xdvriend: The issue here is analogous to the `orthonormalBasis1DM` binding.

gqcpy/include/interfaces.hpp
lelemmen: Yup, that's true! An oversight on my part.

gqcpy/include/interfaces.hpp
lelemmen: Yep, an oversight on my part.

gqcpy/include/interfaces.hpp
xdvriend: This did indeed work! I removed the aliases from the `QCModels` and changed the binding.

gqcpy/include/interfaces.hpp
xdvriend: This can't be rewritten like that. I get the error that there is no matching function for `def`, for all six classes of `StabilityMatrices`(real/complex R/U/GHF).

Screenshot 2020-12-07 at 14 09 42

gqcpy/include/interfaces.hpp
xdvriend: This one can indeed be written like that. I have no idea what causes this one to work and `.internal()` to raise an error.

gqcpy/include/interfaces.hpp
xdvriend: This one gives the same error as `.internal()`.

gqcpy/include/interfaces.hpp
xdvriend: This one also raises the same error when written like that.

xdvriend:
@lelemmen, I've adjusted everything from your suggestions except those things that caused an error to be raised. I've added an explanation of what error was raised to every unresolved conversation above. 

With this, I have re-requested a review.

gqcpy/include/interfaces.hpp
lelemmen: I see, that's probably due to the overloading.

gqcp/include/QCModel/HF/RHF.hpp
xdvriend: You were correct, using a factor `2.0` did the trick!

As for the correctness, the factor 2 comes directly from the article from Pople, where the complex case is tackled as well as the real case.

lelemmen (approved):

PR #858, opened 2020-12-11
Add extra test for Dyson amplitudes

dariatols:
**Short description**
An extra test is implemented with more than 1 electron per spin compound to check whether the value and sign of the Dyson amplitudes are correct. When testing this method with only 1 electron for each spin compound, the sign of each amplitude will always be positive. In the case of multiple `alpha` or `beta` electrons, negative amplitudes can occur.

**Related issues**
https://github.com/GQCG-res/fukui_dyson_outline/issues/43

**Don't forget:**


dariatols:
# K=3 N=4 test system

IMG_0054

dariatols:
It will probably be exactly the same for an alpha electron, but this will be done later this weekend.

lelemmen (approved):
Really nice work with this PR. I like your diligentness about providing a set of hand-written notes and a Jupyter notebook.

Issue #859, opened 2020-12-14
DysonOrbital for SpinUnresolvedONVBasis


dariatols:
**Describe the feature you'd like**
Calculate `DysonOrbital` for `SpinUnresolvedONVBasis` objects.

**Describe what the current code offers in relation to what it lacks**
Now, the code only offers support for `SpinResolvedONVBasis` objects. Since I have written out a [set of notes](https://gqcg-res.github.io/dariatols_knowdes/fukui-matrix.html) for spin-resolved (spin-orbital) and -unresolved (spinor) ONV bases, this should be reflected in the code too.

PR #860, opened 2020-12-14
Dyson orbital for SpinUnresolved ONV basis

dariatols:
**Related issues**
Closes https://github.com/GQCG/GQCP/issues/859.

**Don't forget:**

# K=3 N=2 test system
IMG_0056

# Notebook to verify calculations
Screenshot 2020-12-14 at 15 31 36

### DIFF
--- a/gqcp/include/Processing/Properties/DysonOrbital.hpp
+++ b/gqcp/include/Processing/Properties/DysonOrbital.hpp
@@ -87,7 +87,7 @@ public:
 
         if ((onv_basis1.alpha().numberOfElectrons() - onv_basis2.alpha().numberOfElectrons() != 0) && (onv_basis1.beta().numberOfElectrons() - onv_basis2.beta().numberOfElectrons() != 1)) {
             if ((onv_basis1.alpha().numberOfElectrons() - onv_basis2.alpha().numberOfElectrons() != 1) && (onv_basis1.beta().numberOfElectrons() - onv_basis2.beta().numberOfElectrons() != 0)) {
-                throw std::runtime_error("DysonOrbital::calculateDysonOrbital(LinearExpansion, LinearExpansion): linear_expansion2 is not expressed in a spin-resolved ONV basis with one fewer electron than linear_expansion1.");
+                throw std::runtime_error("DysonOrbital::TransitionAmplitudes(LinearExpansion, LinearExpansion): linear_expansion2 is not expressed in a spin-resolved ONV basis with one fewer electron than linear_expansion1.");
             }
         }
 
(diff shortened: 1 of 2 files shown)


dariatols:
Additional to setting up `DysonOrbital`s for `SpinUnresolvedONVBasis` objects, I have made some adjustments on the naming conventions in the existing `DysonOrbital` class. In the Dyson literature, there is a (quasi) uniform naming convention for the (N-1)- and N-electron states:
- (N-1)-electron states are denoted with letter `I`.
- N-electron states are named `J`.

I suppose that our code should use this convention too, instead of the less informative indices `1` (N-electron system) and `2` ((N-1)-electron system). Do you agree @lelemmen?

gqcp/include/Processing/Properties/DysonOrbital.hpp
```diff
@@ -96,65 +96,124 @@ class DysonOrbital {
...
+
+    /**
+     *  Create a Dyson orbital from the formula for its amplitudes `<N_1|a_p|N>`.
+     * 
+     *  @param linear_expansion_J        The N-electron wave function in a spin-unresolved ONV basis.
+     *  @param linear_expansion_I        The (N-1)-electron wave function in a spin-unresolved ONV basis. It should be expressed in the same orbital basis as the N-electron wave function.
+     *
+     *  @return A Dyson orbital incorporating Dyson amplitudes.
+     */
+    static DysonOrbital<Scalar> TransitionAmplitudes(const LinearExpansion<SpinUnresolvedONVBasis>& linear_expansion_J, const LinearExpansion<SpinUnresolvedONVBasis>& linear_expansion_I) {
```
lelemmen, line 170: I see that you've added the API `TransitionAmplitudes` for given spin-unresolved arguments.

I'm wondering (since we're doing similar things in other parts of the code): do `DysonOrbital`s have fundamentally the same interface (i.e. which methods should you be able to call on them) for spin-resolved and the spin-unresolved cases? If they do, this approach is fine. If they don't, it's better to design a new class.

gqcp/tests/Processing/Properties/DysonOrbital_test.cpp
```diff
@@ -98,3 +98,30 @@ BOOST_AUTO_TEST_CASE(dyson_amplitudes_2) {
     BOOST_CHECK(dyson_coefficients_alpha.isApprox(reference_amplitudes_alpha, 1.0e-06));
     BOOST_CHECK(dyson_coefficients_beta.isApprox(reference_amplitudes_beta, 1.0e-06));
 }
+
```
lelemmen, line 101: You're missing a small explanation of the test here.

lelemmen (commented):
I haven't checked the algorithm, but I trust that your test case is adequate. Thanks for updating the code namings according to widespread notation.

Could you check if currently (or in the near future), `DysonOrbital` should obtain other APIs and accordingly check if we should split up the class due to possible difference between spin-resolved and spin-unresolved behavior?

gqcp/include/Processing/Properties/DysonOrbital.hpp
dariatols: Yes, I think that, other than their (named) constructor, the spin-resolved and -unresolved Dyson orbitals have the same dimension and same behaviour.

dariatols: However, we are using spin-orbitals as an underlying basis for for the `SpinResolvedONVBasis` meaning that we should treat an annihilation on an `alpha` electron differently than an annihilation on a `beta` electron. `TransitionAmplitudes (SpinResolved)` on the other hand does not make a distinction between spin compounds and if we want to describe Dyson **spin-orbitals**, we should separate `alpha` and `beta` contributions.

dariatols: It is not completely clear to me how a Dyson spin-orbital should be defined. [Ortiz2020a](https://doi.org/10.1063/5.0016472) uses a rather vague definition: "Let Dyson spin-orbitals be expanded in terms of `M` orthonormal functions..." (eq. 33). Since eq. (1) in this article has an integral over `x` (space-spin coordinate), it seems that we should perhaps make separate classes for `SpinResolved` and `SpinUnresolved` `DysonOrbital`s.

dariatols: As for the methods that call upon them, they should be quite similar, if not the same. I have not yet seen different approaches for spin-resolved and -unresolved cases in the literature, but maybe I am not familiar enough with the topic yet.

lelemmen (approved):

gqcp/include/Processing/Properties/DysonOrbital.hpp
dariatols: To prevent this discussion from getting out of hand, see the specific issue on this topic https://github.com/GQCG-res/dariatols_knowdes/issues/15.

PR #887, opened 2021-02-18
CC-related improvements

lelemmen:
**Short description**

In this PR, I've updated some CC-related types, and I've implemented a first try for CC DIIS. I'm still experimenting with the implementation, but I'd like to merge some work to develop already, so I can rebuild production containers.

**To do**

### DIFF
--- a/gqcp/include/QCMethod/CC/CCD.hpp
+++ b/gqcp/include/QCMethod/CC/CCD.hpp
@@ -60,7 +60,7 @@ public:
         // Furthermore, the solvers only find the ground state wave function parameters, so the QCStructure only needs to contain the parameters for one state.
         const auto& T2 = environment.t2_amplitudes.back();
 
-        const auto E_electronic_correlation = environment.electronic_energies.back();
+        const auto E_electronic_correlation = environment.correlation_energies.back();
         const QCModel::CCD<Scalar> ccd_parameters {T2};
 
         return QCStructure<GQCP::QCModel::CCD<Scalar>>({E_electronic_correlation}, {ccd_parameters});
--- a/gqcp/include/QCMethod/CC/T2DIIS.hpp
+++ b/gqcp/include/QCMethod/CC/T2DIIS.hpp
@@ -1,0 +1,115 @@
+// This file is part of GQCG-GQCP.
+//
+// Copyright (C) 2017-2020  the GQCG developers
+//
+// GQCG-GQCP is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// GQCG-GQCP is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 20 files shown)


gqcp/include/QCMethod/CC/CCD.hpp
```diff
@@ -60,7 +60,7 @@ class CCD {
         // Furthermore, the solvers only find the ground state wave function parameters, so the QCStructure only needs to contain the parameters for one state.
         const auto& T2 = environment.t2_amplitudes.back();
 
-        const auto E_electronic_correlation = environment.electronic_energies.back();
+        const auto E_electronic_correlation = environment.correlation_energies.back();
```
dariatols, line 63: This name is much clearer and reflects the purpose of this variable in a better way.

gqcp/include/QCMethod/CC/T2DIIS.hpp
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     *  Calculate the accelerated T2 amplitudes and place them in the environment by overwriting the previous T2 amplitudes.
+     * 
+     *  @param environment              The environment that acts as a sort of calculation space.
+     */
+    void execute(Environment& environment) override {
+
+        // Don't do anything if the minimum number of T2 amplitude iterations isn't satisfied.
+        if (environment.t2_amplitude_errors.size() < this->minimum_subspace_dimension) {
+            return;
```
dariatols, line 96: Why don't we return an error if `t2_amplitude_errors.size() <minimum_subspace_dimension` happens?

gqcp/include/QCMethod/CC/T2ErrorCalculation.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     *  Calculate the current T2 error vector and add it to the environment.
+     * 
+     *  @param environment              The environment that acts as a sort of calculation space.
+     */
+    void execute(Environment& environment) override {
+
+        // Read the last two T2 amplitudes iterations and calculate the error as their difference.
+        const auto second_to_last_it = environment.t2_amplitudes.end() - 2;
```
dariatols, line 66: If we only want the last two T2 amplitudes, shouldn't we use `environment.t2_amplitudes.end() - 1` for the second to last element?

dariatols (commented):
Looks great!

xdvriend (approved):
👍

gqcp/include/QCMethod/CC/T2DIIS.hpp
lelemmen: We only want to _activate_ the DIIS when the minimum subspace dimension has been reached.

gqcp/include/QCMethod/CC/T2ErrorCalculation.hpp
lelemmen: No, `.end()` returns an iterator that points to the element _after_ the end of a `vector` (or a `deque`). See also the included example.

Schermafbeelding 2021-02-19 om 14 39 52

dariatols: @lelemmen I was not aware of this, good to know, thanks!

PR #893, opened 2021-03-02
[CMake] Small adjustment for installation ofgqcp and  gqcp_version.

dariatols:
**Short description**
It is a recommended practice to set the header search path during `install()` rather than using the `$<INSTALL_INTERFACE:...>` generator expression on the `gqcp_version` target itself. I have removed the redundant install generator expression in `gqcp` as well, since this header path file was already accounted for in the ìnstall()` command.

**To do**


guacke (approved):
What reference work are you using to refactor the CMake support?

dariatols:
@guacke https://crascit.com/professional-cmake/

PR #901, opened 2021-03-03
[CMake] Bugfix: header files should not be target sources.

dariatols:
**Short description**

The header files should not be included as target source files since these are already included in `target_include_directories`.

**To do**


lelemmen (approved):
Good job!

lelemmen (approved):

PR #920, opened 2021-03-18
Bugfix: GIT_SHA1 variable empty in gqcpy.

dariatols:
**Short description**

Because the `version` library was compiled before the `GIT_SHA1` CMake variable was set, making it empty when importing `gqcpy`.


dariatols:
Screenshot 2021-03-18 at 18 38 51

The `GIT_SHA1` variable is now printed when importing `gqcpy`.

lelemmen:
I see from your screenshot that the code works again. Could you perhaps find a way to unit-test this behavior?

dariatols:
@lelemmen I have included an appropriate test that checks whether or not all version details are empty.

lelemmen (approved):
Looks like a good fix!

PR #922, opened 2021-03-24
Implement LeviCivitaTensor

### LINKED ISSUE #868: Implement LeviCivitaTensor
In #600, we had an implementation of `LeviCivitaTensor`.

Header:
https://github.com/GQCG/GQCP/blob/1e6bcb6a28f2c0e1323484b69d64c193dcfd9f34/gqcp/include/Mathematical/Representation/LeviCivitaTensor.hpp#L27-L66

Source file:
https://github.com/GQCG/GQCP/blob/1e6bcb6a28f2c0e1323484b69d64c193dcfd9f34/gqcp/src/Mathematical/Representation/LeviCivitaTensor.cpp#L18-L65

Test case:
https://github.com/GQCG/GQCP/blob/1e6bcb6a28f2c0e1323484b69d64c193dcfd9f34/gqcp/tests/Mathematical/Representation/LeviCivitaTensor_test.cpp#L18-L79

lelemmen:
**Related issues**

This PR closes #868. More information to be found there.

**To do**


xdvriend (approved):

gqcp/include/Mathematical/Representation/LeviCivitaTensor.hpp
```diff
@@ -0,0 +1,71 @@
...
+namespace GQCP {
+
+
+/**
+ *  The antisymmetric rank-three Levi-Civita tensor.
+ */
+class LeviCivitaTensor {
+private:
+    // The values of the Levi-Civita tensor.
+    Tensor<double, 3> epsilon;
```
AaronDC60, line 33: Why do you use a tensor containing doubles instead of integers, which are smaller in size?
Does this have something to do with possible conversion issues later on?

lelemmen: That's a good question. I guess we could take the scalar type (`double` or `int` or anything else) as a template parameter for this class, so we can accommodate both `LeviCivitaTensor<double>` and `LeviCivitaTensor<int>`.

@AaronDC60 Would you like to implement this changes in this branch?

AaronDC60: Sure, I will give it a try.

lelemmen: You did a great job, @AaronDC60! One thing that I see in the commit history is that your GitHub account does not seem linked to your git commits, as your profile avatar does not appear in the commit timeline. Could you fix this issue?

AaronDC60: I configured my git on the hpc, so this issue should now be fixed.

Issue #865, opened 2020-12-15
Add an ipsocentric CSGT test case


lelemmen:
In order to make sure that our implementation in #498 is correct, we should extensively test the calculations.

Possibly through https://pubs.acs.org/doi/pdf/10.1021/acs.jcim.0c01136 and its supplementing information.

lelemmen:
Update: I'm in contact with Remco to obtain calculated intermediates to check the implementation.

PR #927, opened 2021-03-26
Quantization of current density operator

### LINKED ISSUE #499: Current density 'matrix elements'
In order to support the calculation of the density and current density, we need their 'matrix' elements as required by a second-quantized theory.

We have a general class to support this: `SQOneElectronOperator`, so I think we should proceed in adding first-quantized operators and quantizing them in a spinor basis.

**Additional context**
This also requires #420.

In #600, we had implemented a prototype for the 'current density matrix elements':
https://github.com/GQCG/GQCP/blob/1e6bcb6a28f2c0e1323484b69d64c193dcfd9f34/gqcp/tests/Processing/Properties/properties_test.cpp#L331-L382

### LINKED ISSUE #862: Ipsocentric CSGT for RHF magnetic inducibilities
In #600, we had a prototype for the calculation of ipsocentric CSGT RHF magnetic inducibilities calculated on a grid.

https://github.com/GQCG/GQCP/blob/1e6bcb6a28f2c0e1323484b69d64c193dcfd9f34/gqcp/tests/Processing/Properties/properties_test.cpp#L388-L429

Here, we're doing calculations on a grid, but I think we should find a way to make that API more beautiful.

These calculations were accompanied by the following change in `LinearCombination`

https://github.com/GQCG/GQCP/blob/1e6bcb6a28f2c0e1323484b69d64c193dcfd9f34/gqcp/include/Mathematical/Functions/LinearCombination.hpp#L225-L237

lelemmen:
**Short description**

In this PR, I'm tackling #499 and #862.

For starters:
- I've refactored `ScalarFunction` into `Function`, making sure that any output type is accepted instead of only scalar types;
- `LinearCombination` was renamed to `EvaluableLinearCombination` to reflect its meaning better;
- `EvaluatableScalarRQOneElectronOperator` was renamed to `EvaluableScalarRSQOneElectronOperator` because 'evaluatable' is not a word.

**Related issues**

This PR closes #499, closes #862. A test case (#865) will be included when the currently open PRs #927 and #929 are merged.

**To do**

- [x] Generalize `EvaluableScalarRSQOneElectronOperator` to `EvaluableRSQOneElectronOperator`, making sure that this type can also accommodate the second-quantized current density operator.
- [x] Generalize `EvaluableRSQOneElectronOperator` to its `U` and `G` variants, so we can quantize the electronic density and current density in all R/U/G spinor bases;
- [x] Improve the `calculateDensity` API on `EvaluableRSQOneElectronOperator` and generalize it to also allow the calculation of current densities;


lelemmen:
@AaronDC60 @dariatols I'd really like to tackle some dependent issues as soon as possible. That's why I'm merging this PR without your review. However, please take a look at the implementations in this PR: you can still formulate comments and changes, I'll address them afterwards!

PR #929, opened 2021-03-27
CP(R)HF for magnetic fields

### LINKED ISSUE #376: CPHF for currents
There are two of CP(R)HF equations that have to be solved when calculating response current densities (magnetic inducibilities #498).

- [x] In both cases the 'response force constant' is equal: it is the 'orbital hessian for a purely imaginary response'. In #600, we had implemented a prototype for this Hessian:

https://github.com/GQCG/GQCP/blob/1e6bcb6a28f2c0e1323484b69d64c193dcfd9f34/gqcp/include/QCModel/HF/RHF.hpp#L195-L240

We require the 'response force', i.e. the (negative of) the right-hand side of the linear response equations, for two types of perturbation:
- [x] The perturbation due to the magnetic field;
https://github.com/GQCG/GQCP/blob/1e6bcb6a28f2c0e1323484b69d64c193dcfd9f34/gqcp/tests/Processing/Properties/properties_test.cpp#L249-L265
- [x] The perturbation due to the gauge origin translation;
https://github.com/GQCG/GQCP/blob/1e6bcb6a28f2c0e1323484b69d64c193dcfd9f34/gqcp/tests/Processing/Properties/properties_test.cpp#L274-L299

lelemmen:
**Short description & related issues**

This PR closes #376. I've also re-organized the `QCModel::RHF` APIs.

**To do**


lelemmen:
@AaronDC60 @dariatols I'd really like to tackle some dependent issues as soon as possible. That's why I'm merging this PR without your review. However, please take a look at the implementations in this PR: you can still formulate comments and changes, I'll address them afterwards!

PR #930, opened 2021-03-27
RHF magnetic inducibility

### LINKED ISSUE #862: Ipsocentric CSGT for RHF magnetic inducibilities

lelemmen:
**Related issues**

This PR closes #862. A test case (#865) will be included when the currently open PRs #927 and #929 are merged.

**To do**

### DIFF
--- a/gqcp/include/QCModel/HF/RHF.hpp
+++ b/gqcp/include/QCModel/HF/RHF.hpp
@@ -21,6 +21,7 @@
 #include "Basis/SpinorBasis/OrbitalSpace.hpp"
 #include "Basis/Transformations/RTransformation.hpp"
 #include "DensityMatrix/Orbital1DM.hpp"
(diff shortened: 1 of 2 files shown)


lelemmen:
Closing in favor of #927.

PR #931, opened 2021-03-31
Refactor CMake such that CMAKE_MODULE_PATH no longer has to be set.

guacke:
**Short description**

In this PR I aim to get rid of the requirement that `CMAKE_MODULE_PATH` has to be set manually through command line options passed to CMake.


dariatols (approved):
Nice!

PR #939, opened 2021-04-08
Entanglement implementations

### LINKED ISSUE #785: calculateNDMElement produces wrong results
**Describe the bug**
`calculateNDMElement` in `LinearExpansion.hpp` produces wrong results for some test cases.

**To Reproduce**
```
// Set up an example linear expansion in a spin-unresolved ONV basis.
    const size_t M = 5;
    const size_t N = 2;
    const GQCP::SpinUnresolvedONVBasis onv_basis {M, N};

    const auto linear_expansion = GQCP::LinearExpansion<GQCP::SpinUnresolvedONVBasis>::Random(onv_basis);

    // Check some 1-DM values with calculateNDMElement().
    const auto D_specialized = linear_expansion.calculate1DM();

    BOOST_CHECK(std::abs(linear_expansion.calculateNDMElement({4}, {4}) - D_specialized(4, 4)) < 1.0e-12);  // D(4,4) : a^\dagger_4 a_4
```
`linear_expansion.calculateNDMElement({4}, {4})` will give negative integrals:
```
reference: (6,6) - value: -0.00343283
reference: (7,7) - value: -0.175911
reference: (8,8) - value: -0.0179596
reference: (9,9) - value: -0.134796
```
where the values between brackets are read as `< I | E_{44} | J >` the diagonal element `[4, 4]` of the density matrix which has contributions from 4 combinations of ONVs.

**Expected behavior**
Diagonal elements of the density matrix should have positive contributions since the sign change for an annihilation will be nullified by the sign of the creation operator.

**Additional context**
This was found when I was testing `calculate1DM()` for `SpinUnresolvedONVBasis`. I've used `SpinResolvedONVBasis` with 0 beta electrons now to test my method and it works fine now.

### LINKED ISSUE #875: Hubbard model: single site entropies
For the constrained entanglement research we need a new feature: the ability to calculate the single site entropy in the Hubbard model.

This is not yet implemented in GQCP.

### LINKED ISSUE #878: Pay off technical debt - constrained-entanglement
In order to benefit from the rapid deployment cycles through Docker, I had to push some changes directly to `develop` (or merge using my admin powers) instead of our formal PR&review procedure. Here are the ways we can improve GQCP within the scope of this project:

- [x] Add a test for `LinearExpansion::calculateShannonEntropy`. I changed the internals of the code because they didn't match what a basic Python implementation gave. The error was related to replacing _coefficients_ smaller than `1.0e-18` instead of their square. A good test would be to check if a wave function's Shannon entropy is non-zero for coefficients that are all negative. In the previous implementation, that would have been `0.0` in that case.
- [x] Write `LinearExpansion_bindings` using common interfaces instead of the code duplication that is currently used.
- [x] Introduce a class `AdjacencyMatrix` that facilitates the creation of chains and rings instead of manually setting up the raw adjacency matrices. (#880)

xdvriend:
**Short description**

I fixed the bug in the `calculateNDMElement` API. The old code only took one phase factor into account instead of the separate phase factors from the bra and the ket. 

I also fixed a missing override that caused errors while building.

I used this PR to add the single orbital entropy calculation for two of the 4 types of ONV basis. I'm not sure how to continue with the mutual information, as the `calculateNDMElement` is only enabled for a `SpinUnresolvedONVBasis`, so I'll hold off on that implementation for now.

The Linear expansion python bindings will be refactored using an interface, and an additional test for Shannon entropy will be added.
 
**Related issues**

- closes #785 
- closes #875 
- closes #878 

**To do**

- [x] Implement tests for single orbital entropy
- [x] try to tackle issue #878


dariatols (approved):

xdvriend:
@lelemmen I have requested your review on this PR as everything I set out to address is finished. 

One remark I can already make is concerning the `SeniorityZero` implementation of the `SingleOrbitalEntropy`. I added this implementation to the C++ library, as it is a simplified version of the `SpinResolved` case formulated by Boguslawski. However, when attempting to generate reference data I noticed I couldn't perform a dense CI calculation for `SeniorityZeroONVBases` using gqcpy. I left the implementation in the C++ file for now, but it is not tested. I can either leave it as is, comment it out, or delete it all together. I'll leave that decision up to you.

gqcp/include/QCModel/CI/LinearExpansion.hpp
```diff
@@ -1563,6 +1564,100 @@ class LinearExpansion {
     }
 
 
+    /**
+     * Calculate the single orbital entropy of the orbital at the selected index, from the spin-resolved 1- and two-particle density matrix.
```
lelemmen, line 1568:
```suggestion
     * Calculate the single orbital entropy of the orbital at the selected index, from the spin-resolved 1- and two-particle density matrix. The implementation is based on https://doi.org/10.1002/qua.24832.
```

gqcp/include/QCModel/CI/LinearExpansion.hpp
```diff
@@ -1563,6 +1564,100 @@ class LinearExpansion {
...
+            }
+        }
+
+        // Return the single orbital entropy.
+        return S;
+    }
+
+
+    /**
+     * Calculate the single orbital entropy of the orbital at the selected index, from the spin-resolved 1- and two-particle density matrix.
```
lelemmen, line 1617:
```suggestion
     * Calculate the single orbital entropy of the orbital at the selected index, from the spin-resolved 1- and two-particle density matrix. The implementation is based on https://doi.org/10.1002/qua.24832.
```

lelemmen (approved):
Nice job on the implementations. Also, thanks for refactoring the Python bindings for `LinearExpansion`!

PR #950, opened 2021-04-19
Make entropy test consistent

### LINKED ISSUE #949: SingleOrbitalEntropy test failing on PR's.
**Describe the bug**
The `SingleOrbitalEntropy` test fails on PR #948 without any changes. 

**To Reproduce**
Locally, the test succeeds, but on PR's it sometimes fails.

**Additional context**
I'll try to reproduce the behaviour on develop and look into what could be causing the test to fail.

xdvriend:
**Short description**

Figure out why entropy test sometimes fails.

**Related issues**

closes #949 

**To do**


xdvriend:
I switched the test case to `H4`, for which I also checked the first 5 states and there was no degeneracy. So I believe that this system should be fine. 

I've already requested a review, because the tests succeed locally and hopefully it won't be a problem on GitHub Actions anymore.

lelemmen (approved):
I'm glad we worked out the fix!

PR #951, opened 2021-04-23
Extra bindings for ONVBasis objects.

dariatols:
**Short description**

I have adjusted and added some extra bindings regarding `ONV` representations.

**To do**

### DIFF
--- a/gqcpy/src/ONVBasis/SpinUnresolvedONVBasis_bindings.cpp
+++ b/gqcpy/src/ONVBasis/SpinUnresolvedONVBasis_bindings.cpp
@@ -52,27 +52,21 @@ void bindSpinUnresolvedONVBasis(py::module& module) {
 
         .def(
             "arcWeight",
-            [](const SpinUnresolvedONVBasis& onv_basis, size_t p, const size_t n) {
-                return onv_basis.arcWeight(p, n);
-            },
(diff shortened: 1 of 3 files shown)


gqcpy/src/ONVBasis/SpinUnresolvedONVBasis_bindings.cpp
```diff
@@ -52,35 +52,33 @@ void bindSpinUnresolvedONVBasis(py::module& module) {
...
-        .def(
+        .def_static(
             "calculateDimension",
-            [](const SpinUnresolvedONVBasis& onv_basis, size_t M, const size_t N) {
-                return onv_basis.calculateDimension(M, N);
-            },
+            &SpinUnresolvedONVBasis::calculateDimension,
             py::arg("M"),
             py::arg("N"),
             "Calculate the dimension of a spin-unresolved ONV with M spinors and N electrons.")
```
xdvriend, line 65: Why should this become a `.def_static`?

dariatols: Since its declaration has the `static` keyword `static size_t calculateDimension(const size_t M, const size_t N);`

lelemmen (approved):
Thanks for the clean-up and the additional bindings!

PR #952, opened 2021-04-23
SpinResolved NDM element

### LINKED ISSUE #945: Expand calculateNDMElement to spinResolvedONVBasis
In order to implement the mutual information (#876) we need the ability to calculate 3- & 4-DM elements in a spin resolved ONV basis, which is currently only possible in a `SpinUnresolvedONVBasis`.

xdvriend:
**Short description**

This PR aims to expand the `calculateNDMElement` API to `SpinResolvedONVBasis`. 

**Related issues**
 #945 

**To do**


xdvriend:
As discussed with @lelemmen, we must provide additional implementations before we have the adequate tools to tackle this implementation. Because of that, I'm closing this PR for now.

PR #955, opened 2021-04-30
HF SCF replace_current_XX bindings updated

xdvriend:
**Short description**

This PR updates the aforementioned bindings so that they no longer accept raw matrix representations.

**To do**


gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp
```diff
@@ -103,33 +103,27 @@ void bindQCMethodUHFSCFEnvironmentInterface(Class& py_class) {
          */
 
         .def("replace_current_coefficient_matrices",
```
lelemmen, line 105: Maybe you could also rename this to `replace_current_coefficient_matrix`?

lelemmen (approved):
Look good, I just have a small suggestion in the UHF case.

gqcpy/src/QCMethod/HF/UHF/UHFSCFEnvironment_bindings.cpp
xdvriend: Allright! I edited all UHF API's to reflect the other Hartree-Fock methods.

PR #957, opened 2021-05-05
Notebook showing how to follow GHF internal instabilities

xdvriend:
**Short description**

This PR adds an example on how to use stability analysis to move towards a stable solution in GHF.

**To do**


lelemmen (approved):
Looks good to me. Can you also include it so that it appears on our website?

PR #960, opened 2021-05-10
Implement Python bindings for complex CCD and CCSD

lelemmen:
**Short description**

In this PR, I've made sure that CCD and CCSD can be calculated in magnetic fields.

**To do**


dariatols (approved):
Great!

PR #961, opened 2021-05-10
Implement quantization of S^2

lelemmen:
**Short description**

This PR implements the quantization of the `S^2` operator in a `GSpinorBasis`.

**To do**


xdvriend (approved):
Very clean PR and a really nice addition to GQCP! 

I went through everything, but don't really have any remarks 😄!

PR #966, opened 2021-06-03
Add spinResolved and spinUnresolved classes for an `operator string`

### LINKED ISSUE #964: Spin(Un)ResolvedOperatorString class
In order to more easily work with operator strings in certain CI API's, we need a class to handle the reordering, phase factor calculation, etc. of these operator strings.

xdvriend:
**Short description**

These operator string classes will make working with full CI NDM calculations a lot more feasible and will allow for a clear implementation of the spin resolved NDMElement API.

**Related issues**

closes #964 

**To do**

- [x] Expand the spinUnresolved `isZero()`API to include all cases where an operator string automatically results in a zero value.
- [x] Add a `SpinResolve()`API to the spin resolved class which splits the operator string to it's spinUnresolved components and calculates the correct phase factor in the process.


xdvriend:
@lelemmen, I was pondering the situations where an operator string is inherently equal to zero. But since the operator string classes are defined as either a string of annihilation or a string of creation operators, isn't this always the case once any index appears twice in an operator string? 

We initially talked about the case of adjacent pairs, but I think this is captured within the broader context of the condition I mentioned above.

xdvriend:
@lelemmen,

bar the question/discussion above I have completed the classes as discussed (I believe). Because of that I requested your review.

gqcp/include/ONVBasis/SpinResolvedOperatorString.hpp
```diff
@@ -0,0 +1,169 @@
...
+                        intermediate_phase_factor *= -1;
+                    }
+                }
+            }
+        }
+
+        // The phase factor has been updated in the loops, in order to correspond to a modified operator string with all alpha operators first, followed by all beta operators.
+        // The total operator string is applied on |vac_a>|vac_b>, e.g.: a_a a_a a_b a_b|vac_a>|vac_b>. But we want a_a a_a |vac_a> a_b a_b|vac_b>.
+        // Now we have to apply the fermion anti-commutation rules for eacht time we move |vac_alpha> over a beta operator. This results in p^N_beta.
+        int final_phase_factor = pow(intermediate_phase_factor, beta_index_vector.size());
```
lelemmen, line 162: Moving the (true) vacuum doesn't affect the phase factor since it's empty. I think the correct implementation is the one that ends at line `157`. Could you verify this?

gqcp/include/ONVBasis/SpinResolvedOperatorString.hpp
```diff
@@ -0,0 +1,169 @@
...
+        // Loop over the original pair vector. If the pair contains an alpha spin, add the index to the alpha vector, if the pair contains a beta spin, add the index to the beta vector.
+        for (int i = 0; i < this->index_spin_pairs.size(); i++) {
+            if (this->index_spin_pairs[i].second == GQCP::Spin::alpha) {
+                alpha_index_vector.push_back(index_spin_pairs[i].first);
+            } else {
+                beta_index_vector.push_back(index_spin_pairs[i].first);
+            }
+        }
+
+        // Determine the phase factor needed to make the splitting of the original operator string in its spin components correspond with the fermion anti-commutation rules.
```
lelemmen, line 137:
```suggestion
        // Determine the phase factor needed to make the splitting of the original operator string in its spin components correspond with the fermion anti-commutation rules. In essence, we're counting the number of beta operators that appear in front of every alpha operator.
```

gqcp/tests/ONVBasis/OperatorString_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    // Check whether the indices and spins are stored correctly.
+    BOOST_CHECK_EQUAL_COLLECTIONS(indices.begin(), indices.end(), operator_string_indices.begin(), operator_string_indices.end());
+    BOOST_CHECK_EQUAL_COLLECTIONS(spins.begin(), spins.end(), operator_string_spins.begin(), operator_string_spins.end());
+}
+
+/**
+ *  Check whether the SpinResolvedOperatorString can be correctly `SpinResolved` in its alpha and beta component, with the correct phase factors.
+ */
+BOOST_AUTO_TEST_CASE(test_spin_resolve) {
```
lelemmen, line 132: I don't think this test is correct. I expect `(0a 1b 2a 3b 4a)` to yield a phase factor of `(-1)^3 = -1` because I count 3 beta operators before all the alpha operators:
- `0a` has no beta operators before it, so it is already in the correct place
- `2a` has 1 beta operator before it, so we need 1 anti-commutation: `p=-1`
- `4a` has 2 beta operators before it, so we need 2 anti-commutations: `p=(-1)^2`

So in total, we have `p=(-1)^3 = -1`.

Could you add **two** other tests that you work out by hand to check the implementation?

gqcp/tests/ONVBasis/OperatorString_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+ */
+BOOST_AUTO_TEST_CASE(test_unresolved_constructor) {
+
+    // Initialize the vector containing the indices we want to use for our operator string.
+    std::vector<size_t> indices = {1, 4, 7, 8};
+
+    // Check the constructor.
+    BOOST_CHECK_NO_THROW(GQCP::SpinUnresolvedOperatorString operator_string {indices});
+}
+
```
lelemmen, line 38:
```suggestion


```

gqcp/tests/ONVBasis/OperatorString_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+ *  Check whether the SpinUnresolvedOperatorString constructor works.
+ */
+BOOST_AUTO_TEST_CASE(test_unresolved_constructor) {
+
+    // Initialize the vector containing the indices we want to use for our operator string.
+    std::vector<size_t> indices = {1, 4, 7, 8};
+
+    // Check the constructor.
+    BOOST_CHECK_NO_THROW(GQCP::SpinUnresolvedOperatorString operator_string {indices});
+}
```
lelemmen, line 37: This test doesn't check anything because there are no conditions in `SpinUnresolvedOperatorString` that cause a throw.

gqcp/tests/ONVBasis/OperatorString_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+    // Initialize the vector containing the spin components we want to use for our operator string indices.
+    // However, the dimension of this spin vector doesn't match the index vector dimension and should result in an exception being thrown.
+    std::vector<GQCP::Spin> spins_wrong = {GQCP::Spin::alpha, GQCP::Spin::beta, GQCP::Spin::alpha};
+
+    // Check the constructor.
+    BOOST_CHECK_THROW(GQCP::SpinResolvedOperatorString operator_string_wrong(indices_wrong, spins), std::invalid_argument);
+    BOOST_CHECK_THROW(GQCP::SpinResolvedOperatorString operator_string_wrong_2(indices, spins_wrong), std::invalid_argument);
+    BOOST_CHECK_NO_THROW(GQCP::SpinResolvedOperatorString operator_string_correct(indices, spins));
+}
+
```
lelemmen, line 63:
```suggestion


```

gqcp/tests/ONVBasis/OperatorString_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+    // Initialize the SpinUnresolvedOperatorString.
+    const auto operator_string = GQCP::SpinUnresolvedOperatorString {indices};
+
+    // Get the indices from the operator string.
+    const auto operator_string_indices = operator_string.operatorIndices();
+
+    // Check whether the indices are stored correctly.
+    BOOST_CHECK_EQUAL_COLLECTIONS(indices.begin(), indices.end(), operator_string_indices.begin(), operator_string_indices.end());
+}
+
```
lelemmen, line 81:
```suggestion


```

gqcp/tests/ONVBasis/OperatorString_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+/**
+ *  Check whether the SpinUnresolvedOperatorString stores its information correctly.
+ */
+BOOST_AUTO_TEST_CASE(test_unresolved_indices) {
+
+    // Initialize the vector containing the indices we want to use for our operator string.
+    std::vector<size_t> indices = {1, 4, 7, 8};
+
+    // Initialize the SpinUnresolvedOperatorString.
+    const auto operator_string = GQCP::SpinUnresolvedOperatorString {indices};
```
lelemmen, line 73:
```suggestion
    const GQCP::SpinUnresolvedOperatorString operator_string {indices};
```
In your statement, you're doing a construction and then an assignment. In this statement, you'll only be doing the construction. C++ isn't Python!

gqcp/tests/ONVBasis/OperatorString_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    // Initialize the vector containing the indices we want to use for our operator string.
+    // One vector does contain duplicate indices, one does not.
+    std::vector<size_t> indices_1 = {1, 4, 7, 8};
+    std::vector<size_t> indices_2 = {1, 4, 7, 1};
+
+
+    // Initialize the SpinUnresolvedOperatorStrings corresponding to the indices.
+    const auto operator_string_1 = GQCP::SpinUnresolvedOperatorString {indices_1};
+    const auto operator_string_2 = GQCP::SpinUnresolvedOperatorString {indices_2};
```
lelemmen, line 95:
```suggestion
    const GQCP::SpinUnresolvedOperatorString operator_string_1 {indices_1};
    const GQCP::SpinUnresolvedOperatorString operator_string_2 {indices_2};
```

gqcp/tests/ONVBasis/OperatorString_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    // Use the `isZero()` API to check whether the operator string will always result in a zero value.
+    const auto operator_string_1_zero_check = operator_string_1.isZero();
+    const auto operator_string_2_zero_check = operator_string_2.isZero();
+
+    // Compare the result of the `isZero()` check with the expected boolean value (false for the operator string without duplicate indices, true for the one with duplicate indices).
+    BOOST_CHECK_EQUAL(operator_string_1_zero_check, false);
+    BOOST_CHECK_EQUAL(operator_string_2_zero_check, true);
```
lelemmen, line 103:
```suggestion
    // Compare the result of the `isZero()` check with the expected boolean value (false for the operator string without duplicate indices, true for the one with duplicate indices).
    BOOST_CHECK_EQUAL( operator_string_1.isZero(), false);
    BOOST_CHECK_EQUAL(operator_string_2.isZero(), true);
```
No need to make things so verbose.

gqcp/tests/ONVBasis/OperatorString_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+BOOST_AUTO_TEST_CASE(test_resolved_indices_spins) {
+
+    // Initialize the vector containing the indices we want to use for our operator string.
+    std::vector<size_t> indices = {1, 4, 7, 8};
+
+    // Initialize the vector containing the spin components we want to use for our operator string indices.
+    std::vector<GQCP::Spin> spins = {GQCP::Spin::alpha, GQCP::Spin::beta, GQCP::Spin::alpha, GQCP::Spin::beta};
+
+    // Initialize the SpinUnresolvedOperatorString.
+    const auto operator_string = GQCP::SpinResolvedOperatorString(indices, spins);
```
lelemmen, line 118:
```suggestion
    const GQCP::SpinResolvedOperatorString operator_string {indices, spins};
```

lelemmen (changes requested):
I think the implementation for the phase factor in the spin-resolved case is wrong.

lelemmen:
> @lelemmen, I was pondering the situations where an operator string is inherently equal to zero. But since the operator string classes are defined as either a string of annihilation or a string of creation operators, isn't this always the case once any index appears twice in an operator string?
> 
> We initially talked about the case of adjacent pairs, but I think this is captured within the broader context of the condition I mentioned above.

Yes, since we're considering only pure creation or annihilation strings, you are correct!

gqcp/include/ONVBasis/SpinResolvedOperatorString.hpp
xdvriend: You are correct, this one is on me. In the notes from our meeting we split the string in `D_a` and `I_a` and while implementing I mistook moving `I_a` part over the beta operators for moving the vacuum. 

As it is empty, moving the vacuum indeed doesn't change the phase factor. I'll fix this and adjust the documentation where necessary.

gqcp/tests/ONVBasis/OperatorString_test.cpp
xdvriend: Indeed, this is due to the error I made concerning the vacuum. 

While implementing, I printed the intermediate phase factor values and this did result in what you are describing here: `(-1)^3`. The `p=1` I was testing for comes from the factor `^2` I added for moving the vacuum. I`ll remove the power and add additional tests to support this.

gqcp/include/ONVBasis/SpinUnresolvedOperatorString.hpp
```diff
@@ -0,0 +1,107 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with GQCG-GQCP.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+
```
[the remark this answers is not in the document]
xdvriend: I added `<vector>` to both classes because they both use `std::vector` and I also added `<iterator>` to the spinUnresolved class because of the `std::unique` iterator in the `isZero()` method. I believe this should cover all missing headers.

gqcp/include/ONVBasis/SpinResolvedOperatorString.hpp
xdvriend: @lelemmen I just realized my explanation above wasn't entirely clear. The `I_a` in our notes represented the alpha creation operators acting on the alpha vacuum, which in total represented the ONV on which the operator string should be acting. 

When moving these alpha operators across the beta operators of the spin resolved operator string, we have to take the  anti-commutation rules into account. However, this is context dependant (since it depends on the ONV on which the operator string acts), and should not be factored in in this implementation, as here, we focus purely on the operator string itself. 

Long story short, I did indeed make a mistake here & you are correct that what this API has to do is implemented up until line 157, as moving the vacuum has no effect. However, when using this class in for example the future `calculateNDMElement` API's, we still have to take the anti commutation rules of moving the `I_a^dagger` operators.

Hope this clears things up!

lelemmen: Yes, the extra anti-commutation relations resulting from `I_a^dagger` should be accounted for, but (as you said) not in this class.

gqcp/include/ONVBasis/SpinUnresolvedOperatorString.hpp
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Check whether the operator string in question will result in zero when applied to the wave function.
+     *
+     * Note: There are different cases when an operator string will result in a zero value. This method checks all of them.
+     */
+    bool isZero() const {
+
+        // Check if the same index appears more than once in the operator string.
+        // Start by initializing a copy of the index vector associated with the operator string.
+        auto index_vector = this->operatorIndices();
```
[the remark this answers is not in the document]
xdvriend: I didn't change this as the `sort` and `std::unique` functions change the vector and using `const auto& index_vector` would cause the compilation to fail.

gqcp/include/ONVBasis/SpinResolvedOperatorString.hpp
```diff
@@ -0,0 +1,169 @@
...
+// along with GQCG-GQCP.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+
+#include "ONVBasis/SpinUnresolvedOperatorString.hpp"
+#include "QuantumChemical/Spin.hpp"
+#include "QuantumChemical/SpinResolved.hpp"
+
+#include <math.h>
```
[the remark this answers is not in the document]
xdvriend: As the `pow()` functions was removed from the implementation, I removed this header all together.

xdvriend:
@lelemmen I have adjusted everything from your suggestions. Some of my adjustments have been clarified in the conversations above, in order to avoid confusion as to why I did something a certain way. 

With this I believe this PR is ready for review again.

dariatols (approved):
👍🏻

lelemmen:
@xdvriend This PR totally slipped out of my mind. I apologize. I think this one is good to go.

lelemmen (approved):

xdvriend:
> @xdvriend This PR totally slipped out of my mind. I apologize. I think this one is good to go.

No problem at all! Thanks for the review! 😄

PR #970, opened 2021-06-13
Fix GA conda uploads to Anaconda. Closes #967.

### LINKED ISSUE #967: Github Actions does not upload to Anaconda
**Describe the bug**
Our Conda builds are not uploaded to Anaconda on push to develop.

guacke:
Anaconda tokens have an expiration date and the Anaconda Client does **not** warn when this token has expired, nor does it notify that the upload has failed. I have updated the token and I am now testing if uploading again works. Update: updating the token indeed fixes the upload.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@
 #   - the gqcpy Python bindings to the C++ library (optional)
 
 cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
-project(GQCP VERSION 1.0.0 LANGUAGES CXX)
+project(GQCP VERSION 0.1.0 LANGUAGES CXX)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)  # make CMake look into the ./cmake/ folder for configuration files
 
 # Look for the dependencies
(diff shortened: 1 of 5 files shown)


guacke:
Pre-merge clean-up: 
- [ ] Do not upload pull request builds to anaconda.

dariatols (approved):

CMakeLists.txt
```diff
@@ -7,7 +7,7 @@
 #   - the gqcpy Python bindings to the C++ library (optional)
 
 cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
-project(GQCP VERSION 0.0.1 LANGUAGES CXX)
```
lelemmen, line 10: I realized during the weekend that releasing a `v1.0.0` might not be the best idea yet. There still remain some old APIs (such as those related to AP1roG, response classes and orbital optimizers) that, in my opinion, should be improved  beforehand. Committing ourselves to a `v1.0.0` would mean throwing away the freedom we have to break APIs in `v0.x.y` versions.

I do realize that there is a minor inconsistency in our versioning at the moment, so perhaps we could refer to the 'paper' release as `v0.1.0` and from then on rigorously use the semantic versioning system?

guacke: 👍

lelemmen (approved):

PR #980, opened 2021-09-03
Add quantization of `FQMolecularPauliHamiltonian` in an `USpinorBasis`.

lelemmen:
**Short description**

This PR adds the quantization of `FQMolecularPauliHamiltonian` in a `USpinOrbitalBasis` composed of `LondonGTOShell`s.

PR #982, opened 2021-09-08
Bugfix for conda CI.

dariavh:
**Short description**

This PR tries to solve the include errors in PR #980 and makes sure that the conda CI works for pushes on `develop` **and** on pull requests.

**Related issues**

#980.

**To do**

### DIFF
--- a/.github/workflows/ci-cd-conda.yml
+++ b/.github/workflows/ci-cd-conda.yml
@@ -54,6 +54,8 @@ jobs:
       - name: Build conda package
         if: github.event_name == 'pull_request'
         run: |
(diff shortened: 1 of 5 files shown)


dariavh:
@lelemmen @guacke The conda builds in #980 did not work due to a GCC compiler upgrade to version 11 recently. After including the right libraries in some header files, the CI is up and running again.

During the fix, I noticed that the conda build **always** used the last commit on `develop` as a starting point instead of the HEAD on the current branch, meaning that conda builds on pull requests were not testing the pull requests themselves. This is solved in this PR as well.

guacke (approved):
👍

PR #983, opened 2021-09-08
Create Docker image locally.

dariavh:
**Short description**

Look at the possibility to build the Docker image locally as a tar file and use it on the cluster. (see https://stackoverflow.com/questions/52738888/converting-docker-tar-file-into-a-singularity-image)

**To do**

### DIFF
--- a/.github/workflows/docker-gqcp-deploy.yml
+++ b/.github/workflows/docker-gqcp-deploy.yml
@@ -38,9 +38,3 @@ jobs:
         file: docker/Dockerfile
         push: true
         tags: gqcg/gqcp:latest
-        outputs: type=docker,dest=gqcp.tar
-    - name: Upload artifact
-      uses: actions/upload-artifact@v2
-      with:
-        name: gqcp
-        path: gqcp.tar
--- a/.github/workflows/docker-gqcp-deploy.yml
+++ b/.github/workflows/docker-gqcp-deploy.yml
@@ -38,3 +38,9 @@ jobs:
         file: docker/Dockerfile
         push: true
         tags: gqcg/gqcp:latest
+        outputs: type=docker,dest=gqcp.tar
+    - name: Upload artifact
+      uses: actions/upload-artifact@v2
+      with:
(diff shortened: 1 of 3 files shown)


dariavh:
@guacke This PR generates a `tar` file for a Docker image as an artifact, but this cannot be used on the cluster as it gives the following error 
```
FATAL:   You must be the root user, however you can use --remote or --fakeroot to build from a Singularity recipe file
```
when I perform
```
singularity build local_gqcp.sif gqcp.tar
```
or even
```
FATAL:   could not use fakeroot: no mapping entry found in /etc/subuid for vsc42645
```
when I use `--fakeroot`
```
singularity build --fakeroot local_gqcp.sif gqcp.tar
```


EDIT: Mail to HPC helpdesk has been sent.

dariavh:
@guacke This does not seem to be possible on the HPC as the--fakeroot option is broken.

8550FF83-419D-4441-A559-C8E3E029087D

guacke:
@dariatols On which cluster did you try to build the Singularity image? 

Maybe we should look into providing a separate [Singularity Definition File](https://sylabs.io/guides/3.7/user-guide/definition_files.html) and connecting to the [Singularity Container Services](https://cloud.sylabs.io/home) (although this _seemingly_ introduces redundant code with respect to the Dockerfile).

dariavh:
@guacke I tried it on victini. I'll take a look at your alternative approach tomorrow.

The helpdesk suggests to build the .sif file locally, and then move it to the cluster. Is that a possibility as well? I could try to build it on GitHub Actions and create an artifact for the sif file.

dariavh:
Screenshot 2021-09-09 at 13 10 50


@guacke The workflow `Singularity image` provides an artifact called `gqcp.sif`. If you move this to your scratch on the HPC, you can shell into the image! I will clean up this PR and assign you as reviewer.

EDIT: The Singularity image build departs from the version on DockerHub. I will change this in the next commit(s) to let the build start from a local tar file.

PR #986, opened 2021-09-14
Initiate Docker build when Dockerfile is changed

dariavh:
The `Dockerfile` is now in a subfolder called `docker`.


xdvriend (approved):
👍

PR #989, opened 2021-09-16
Add unit tests for complex Hartree-Fock

xdvriend:
**Short description**

Due to a bug found by @lelemmen it was necessary to double check the remaining einsum API's as well as add unit tests for the complex Hartree-Fock methods. 

Another bug was found while doing this, which caused the complex RHF method to not converge. The initial contraction pairs (from before the einsum refactor) were `(3, 0) (2, 1)` for the direct matrix and `(1, 0) (2, 1)`for the exchange matrix, which translates to `ijkl,lk->ij` and `ijkl,jk->il` in einsum notation.  

However, from comparison with UHF and GHF and double checking the theory I noticed that the einsum notation should be `ijkl, kl->ij` and `ijkl, kj->il`, or in contraction pairs `(2, 0) (3, 1)` and `(2, 0) (1, 1)`. 

After applying this change, the complex RHF method converged to the expected lower lying state that I double checked with my python code and stability analysis.

### DIFF
--- a/gqcp/include/QCModel/HF/RHF.hpp
+++ b/gqcp/include/QCModel/HF/RHF.hpp
@@ -728,9 +728,9 @@ public:
 
         // To calculate G, we must perform two double contractions:
         //      1. (mu nu|rho lambda) P(lambda rho),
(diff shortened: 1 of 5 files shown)


lelemmen (approved):
Thanks for taking the time to add these tests.

Issue #990, opened 2021-09-27
Implementation for Hubbard entanglements


dariavh:
In order to calculate the entropy and the orbital interaction, we need to be able to calculate reduced density matrices (see eq. 4c in [Mund2009](https://doi.org/10.1103/PhysRevB.79.245130)) and use their eigenvalues. This issue will collect all necessary steps for the implementation.

- [x] Divide wavefunction into two subsystems as written down by [Mund2009](https://doi.org/10.1103/PhysRevB.79.245130).<img width="180" alt="Screenshot 2021-09-27 at 11 55 09" src="https://user-images.githubusercontent.com/33431368/134887102-634f369c-026c-4927-b3f0-69951e940d98.png">
See #991.
- [x] Construct a reduced density matrix for the subsystem of interest. <img width="151" alt="Screenshot 2021-09-27 at 11 58 45" src="https://user-images.githubusercontent.com/33431368/134887641-1364a0ea-3e04-405c-bb58-885a9e19fd6d.png">
See #994.
- [x] Calculate the von Neumann entropy using the eigenvalues/trace of the reduced density matrix.

PR #994, opened 2021-10-11
Reduced density matrix calculations

dariavh:
**Short description**

This PR provides a notebook (and extra bindings) to calculate the reduced density matrix according to equation 7 in [Boguslawski2014](https://arxiv.org/abs/1409.8017).

Additionally, it changes the name of the `schmidtDecomposition` method since the use of this name is not correct.

**Related issues**

#990 

**To do**


xdvriend (approved):
👍

PR #1009, opened 2021-10-21
Add stl.h library for automatic conversions.

dariavh:
**Short description**

Add automatic conversion for `std::vector<>`.

**To do**


xdvriend (approved):
👍

PR #1017, opened 2021-12-06
Optimization for calculateSystemOrbitalRDM.

dariavh:
**Short description**

As the method `calculateSystemOrbitalRDM` becomes computationally expensive for larger systems, `GQCP` could benefit from an optimization where matrix elements are zero when
Screenshot 2021-12-06 at 14 30 42

(see [Rissler2005](https://doi.org/10.1016/j.chemphys.2005.10.018))

**To do**


xdvriend (approved):
Looks good to me!

PR #1020, opened 2022-02-02
Adjust Python bindings for LinearExpansion.

dariavh:
**Short description**

We cannot return a `GQCP::Tensor<double, 2>` type to Python, this should be type casted to a `GQCP::Matrix<double>` type.

**To do**


xdvriend (approved):
👍

PR #1021, opened 2022-02-22
Feature/sylabs

guacke:
**Short description**

This PR will focus on providing Sylabs.io cloud support for Singularity containers.

**Related issues**

When converting our Docker files to Singularity containers, we encounter difficulties with hanging kernels when using np.linalg.


guacke:
The Singularity SIF can be built using GA. Replacing upload to sylabs.io with upload to Github Container Registry.

PR #1022, opened 2022-02-22
Following internal instabilities

xdvriend:
**Short description**

In line with the revision of the spin phase transitions paper, there is a need for following internal Hartree-Fock instabilities. This PR will provide this functionality from within the `StabilityMatrices` classes. This can later be revised into the correct `OrbitalRotationGenerator` classes and cleaner API's.

**To do**

- [x] Provide functionality to follow internal instabilities in real GHF.
- [x] Provide functionality to follow internal instabilities in complex GHF.
- [x] Provide functionality to follow internal instabilities in real UHF.
- [x] Provide functionality to follow internal instabilities in complex UHF.
- [x] Implement python bindings for all of the above methods.


xdvriend:
The test for the complex case does not compare the resulting transformation matrix since the diagonalization of the complex matrix results in different eigenvectors for `NumPy`and `Eigen`. The eigenvalues are the same, and in both cases the transformation leads to a stable solution.

I have no idea why the build fails here, it works locally. I'll look into it when I add the UHF cases as well.

EDIT: Builds now working. Test probably failing due to H3 degeneracy. Could also explain complex situation.

PR #1023, opened 2022-03-08
fix the stability threshold in the python bindings

xdvriend:
**Short description**

The stability threshold in the python bindings was set to `1e-5` instead of `-1e-5` causing some methods to return a false statement about stability when used in python.


dariavh (commented):
👍🏻👍🏻👍🏻

dariavh (approved):

lelemmen:
Ooh, that's a nasty one to track down. Good job! 😃

PR #1030, opened 2022-08-24
OneZero UNOCI calculation gives an Eigen error

xdvriend:
**Short description**

A UNOCI calculation with one zero overlap value leads to an eigen error at the moment.


xdvriend:
After rebuilding everything from scratch as mentioned [here](https://github.com/GQCG/GQCP/pull/1031), the `crash_test.cpp` fails with the following error:

`Intel MKL FATAL ERROR: Cannot load /usr/local/miniconda3/lib/libcint.so.`

xdvriend:
I have tracked the original error down to the coDensity matrix. I'll further investigate what exactly goes wrong here.

xdvriend:
I have figured out the origin of the bug. The alpha part has two occupied orbitals, and thus, the dimension of the biorthogonal expansions is (3, 2). The beta part has only one occupied orbital, as such, the biorthogonal expansions have dimension (3, 1). 

However, when a ULowdinPairingBasis calculates a co-density, it calculates the alpha and beta part for the given index. Since the index for the alpha zero overlap is `1` the beta part cannot be calculated and gives an error.

This error originates since the pair vector containing the zero overlap indices, paired with their respective spins, adds the alpha zero overlaps before the beta zero overlaps. I believe, adding them ordered by their respective index will solve these errors.

Tagging @guacke.

xdvriend:
I have everything succeeding on my own system (where I was previously able to reproduce the bug we encountered). I added new tests for these one and two zero edge cases. 

If everything succeeds here, and you agree (@guacke) I will merge this PR.

guacke:
@xdvriend Could first merge PR #1031 with this branch? Conda errors in this branch, because an incorrect version of libcint is currently being used. Note that to resolve the errors mentioned in PR #1031, you have to use the build instructions as given in the `conda-ci-cd` Github Action.

xdvriend:
> @xdvriend Could first merge PR #1031 with this branch? Conda errors in this branch, because an incorrect version of libcint is currently being used. Note that to resolve the errors mentioned in PR #1031, you have to use the build instructions as given in the `conda-ci-cd` Github Action.

I merged the PR's together and will await to see whether or not the builds succeed. I believe I adjusted the builds locally to match everything so that aspect should be fine. If the builds succeed, I'll merge this PR with develop.

PR #1031, opened 2022-08-27
Devops/enable conda tests

guacke:
**Short description**

While the build-tests are failing in PR #1030, the conda builds seem ok. However, this is because the test are apparently disabled in the conda build. I will re-enable them in this PR.

Tagging @xdvriend .

### DIFF
--- a/conda/build.sh
+++ b/conda/build.sh
@@ -19,6 +19,7 @@ if [ `uname` == Darwin ]; then
         -DCMAKE_CXX_FLAGS="${CXXFLAGS} ${OPTS}" \
         -DCMAKE_VERBOSE_MAKEFILE=TRUE \
         -DBUILD_PYTHON_BINDINGS=TRUE \
+        -DBUILD_TESTS=TRUE \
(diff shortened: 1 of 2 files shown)


guacke:
@xdvriend Conda reports an error in GLowdinPairingBasis_test.cpp.

xdvriend:
> @xdvriend Conda reports an error in GLowdinPairingBasis_test.cpp.

I'v been looking into it, but I can't reproduce this error. (completely deleted and reinstalled GQCP, using the exact same settings as the GitHub actions build) It is also strange that the error would occur while building the `GLowdinPairingBasis_test.cpp`, as I've never had any issues with calculations in a generalised basis. 

Do all Conda builds raise this error? Or are others just being cancelled, as we've sometimes had issues with that as well, which are often solved by just restarting the build.

xdvriend:
Since the conda builds are being cancelled on [other PR's](https://github.com/GQCG/GQCP/pull/1030) as well, I believe this might be the same thing happening here. I don't think `GLowdinPairingBasis_test.cpp` is causing the builds to fail.

PR #1042, opened 2023-03-21
Domain python bindings

xdvriend:
**Short description**

The new domain functionalities should be bound to python API's.

**Related issues**

[This issue in GQCC.](https://github.com/GQCG-res/GQCConstraints/issues/54#event-8063738641)


dariavh (approved):
Looks great!

gqcp/include/Domain/DiscreteDomain.hpp
```diff
@@ -120,7 +120,7 @@ class DiscreteDomain:
      *
      *  @return     The i-th element with values 0 or 1 depending on whether the i-th element is in the domain.
      */
-    bool operator()(const size_t i) const { return this->domain[i]; }
+    bool operator[](const size_t i) const { return this->domain[i]; }
```
dariavh: I chose the `()` method since it is in line with the style that `GQCP` currently uses in `GQCP::Matrix` classes and derivatives thereof.

xdvriend: Right, my bad! I changed it in order to try something with the python bindings, but it didn't work & I forgot to change it back. I'll adjust this before merging.

Issue #1043, opened 2023-03-29
Docker build hanging


dariavh:
**Describe the bug**
The Docker build hangs when solving the environment, which causes the build to exceed the time limit of 6 hours. See commits https://github.com/GQCG/GQCP/commit/b9625021b262e93cee87fd042caa7b19f22c531c, https://github.com/GQCG/GQCP/commit/8ff5d54f8d617d700a7ea89563e03b789ff8f881 and https://github.com/GQCG/GQCP/commit/0b8216329c6ef2bd8c344c255fd3e8b87074e11f.

**To Reproduce**
Pushing to develop will cause the Docker build to fail.

**Screenshots and logs**
If applicable, add screenshots or logs containing warnings and errors to help explain your problem.

**Operating system**
 - OS: ubuntu 
 - Version 20.04

PR #1044, opened 2023-03-29
Bugfix Docker build.

dariavh:
**Short description**

The latest Docker builds hang when solving the environment, causing the build to fail.

**Related issues**

#1043 

**To do**


dariavh:
Closed in favour of #1046.

PR #1046, opened 2023-04-05
Upgrade dependencies to fix conda search for dependencies

dariavh:
**Short description**

The latest Docker builds hang when solving the environment, causing the build to fail.

**Related issues**

#1043 

**To do**

- [x] Bump dependencies to the latest version (if possible).
- [x] Check compatibility of `GQCP` code with upgraded dependencies.
- [x] Adjust Docker containers `GQCP-dev` and `GQCP`.


dariavh:
@guacke @xdvriend At the moment, all builds are successful except for the conda macOS build, which appears to halt randomly during the build of `DoubleLondonHermiteCoulombIntegral.cpp`, without a clear error message. Despite attempting to increase the macOS version to 12, the issue still persists. The conda installer is able to detect all necessary dependencies, so I am currently unsure of how to proceed with solving this problem. Should we disable the macOS conda build?

PS: The Linux conda build does not start when the macOS conda build fails, but as you can see, the Linux conda build succeeds when disabling the macOS version.

xdvriend (approved):
Great work on these dependencies!

PR #1048, opened 2023-04-17
Implement partitions over Domains.

dariavh:
**Short description**

This PR will provide new features regarding the partition of domains, electron numbers and orbitals over the `Domain`s defined by `GQCP`.

**Related issues**
Closes #1049.

**To do**

- [x] (if applicable, when creating new source files): don't forget to update the Doxygen file, the collective `gqcp.hpp` include header and the CMake files
- [x] Implement `SimplePartition` as a base class for all partitions.
- [x] Implement `DomainPartition` as a base class for partitions of domains.
- [x] Implement `DiscreteDomainPartition` that divides a discrete set of orbitals into domains.
- [x] Generalize functionality `overlapWithONV` from `HubbardDomain` to `DiscreteDomain`.
- [x] Implement `ONVPartition` to create new ONVs for each (discrete) domain separately.
- [x] Implement `ElectronPartition` (spin-unresolved and spin-resolved) to specify the electron occupation in each domain.
- [x] Write tests.
- [x] Write Python bindings.

PR #1051, opened 2023-06-05
Refactor OrbitalRDM calculation according to DomainPartition.

dariavh:
**Short description**

In the current version of `calculateSystemOrbitalRDM`, the user has to provide the ONVs of the system and environment. Due to the new functionalities introduced in #1048, I have refactored this method such that the user now only needs to provide the domain partition to partition the `LinearExpansion` into subsystems.


xdvriend (approved):
Looks good to me!

Issue #1091, opened 2025-10-10
Coulomb and Exchange matrix python bindings


ThiboVE:
**What is missing**
Currently GQCPy has no way of accessing the Coulomb and Exchange matrices of a HF method

**Solution**
Implement the python bindings:

In [GQCP/gqcpy/include/interfaces.hpp](https://github.com/GQCG/GQCP/blob/develop/gqcpy/include/interfaces.hpp#L1221), add

```C
.def(
            "calculateScalarBasisDirectMatrix",
            &Type::calculateScalarBasisDirectMatrix,
            py::arg("density_matrix"),
            py::arg("hamiltonian"),
            "Return the Coulomb matrix")

.def(
            "calculateScalarBasisExchangeMatrix",
            &Type::calculateScalarBasisExchangeMatrix,
            py::arg("density_matrix"),
            py::arg("hamiltonian"),
            "Return the Exchange matrix")
```

To the "bindQCModelHartreeFockInterface" class

This directly implements the bindings for all HF methods.

To make this work, however, the RHF implementation in C++ needs to be changed. J and K are calculated directly in the "calculateScalarBasisFockMatrix" function, while they need to be put into their own function.

emielvb:
@ThiboVE it appears something went wrong, as all files have been changed. Instead, I will open a new PR and ensure only those files that _have_ been changed, show up in the diff, and will then work from there to close #1091.

**Which files did you yourself bring _actual_ changes to?**
From #1091 i see you changed a binding file [GQCP/gqcpy/include/interfaces.hpp](https://github.com/GQCG/GQCP/blob/develop/gqcpy/include/interfaces.hpp#L1221), are there any other files you changed, e.g. with cpp implementations of the new J and K functions, any headers, etc?
Please list them here, so that I can build upon your work.

ThiboVE:
@emielvb all issues I have made were proposals for to be implemented changes/features. The changes I actually made were restricted to the RHF implementation in [gqcp/include/QCModel/HF/RHF.hpp] as this is required to realize the python bindings in this issue (see issue #1095)

PR #1093, opened 2025-10-15
separate K and J from Fock matrix calculation in RHF

ThiboVE:
**Short description**

Separate the calculation of the Coulomb and Exchange matrix from the function that calculates the Fock matrix in RHF.

**Related issues**

[Coulomb and Exchange matrix python bindings ](https://github.com/GQCG/GQCP/issues/1091)

**To do**

- [x] Create a function for caluclating the Coulomb matrix in RHF in cpp code
- [x] Create a function for calculating the Exchange matrix in RHF in cpp code


emielvb:
Superseded by #1096

PR #1096, opened 2025-10-29
Implement functions to separate coulomb and exchange energy contributions in RHF

### LINKED ISSUE #1091: Coulomb and Exchange matrix python bindings

### LINKED ISSUE #1095: Coulomb and Exchange function in RHF (C++ code) are missing
**Context**
- To realize issue #1091, all HF methods need to have distinct functions for calculating the Coulomb and Exchange matrix in scalar basis.
- Current RHF implementation (in [gqcp/include/QCModel/HF/RHF.hpp](https://github.com/GQCG/GQCP/blob/develop/gqcp/include/QCModel/HF/RHF.hpp)) calculates the Coulomb and Exchange matrix directly in the function of the Fock matrix.

**Problem**
Implementing the bindings stated in issue #1091 will lead to an error when calling these bindings for RHF as there are no Coulomb and Exchange matrix functions in the C++ code.

**Solution**
Remove the construction of the Coulomb and Exchange matrix from the Fock matrix construction and assign them their own function.

emielvb:
**Supersedes #1093**.
Closes #1095 and #1091.

**Short description**

Separate the calculation of the Coulomb and Exchange matrix from the function that calculates the Fock matrix in RHF.

**Related issues**

[Coulomb and Exchange matrix python bindings ](https://github.com/GQCG/GQCP/issues/1091)

**To do**

- [x] Create a function for caluclating the Coulomb matrix in RHF in cpp code
- [x] Create a function for calculating the Exchange matrix in RHF in cpp code

### DIFF
--- a/gqcp/include/QCModel/HF/RHF.hpp
+++ b/gqcp/include/QCModel/HF/RHF.hpp
@@ -738,6 +738,47 @@ public:
         return RHF<Scalar>::calculateScalarBasis2DM(this->expansion(), N);
     }
 
+    /**
+     *  Calculate the RHF direct (Coulomb) operator.
+     *
+     *  @param P                    The RHF density matrix expressed in the underlying scalar orbital basis.
+     *  @param sq_hamiltonian       The Hamiltonian expressed in the (same) underlying scalar orbital basis.
+     *
+     *  @return The RHF direct (Coulomb) operator.
+     */
+    static ScalarRSQOneElectronOperator<Scalar> calculateScalarBasisDirectMatrix(const Orbital1DM<Scalar>& D, const RSQHamiltonian<Scalar>& sq_hamiltonian) {
+
+        // Get the two-electron parameters.
+        const auto& g = sq_hamiltonian.twoElectron().parameters();
+
+        // To calculate J, we must perform a double contraction:
+        //      (mu nu|rho lambda) P(lambda rho),
+        const auto J = g.template einsum<2>("ijkl,kl->ij", D.matrix()).asMatrix();
+
+        return ScalarRSQOneElectronOperator<Scalar> {J};
+    };
+
+
+    /**
+     *  Calculate the RHF exchange operator.
+     *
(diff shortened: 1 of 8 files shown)


emielvb:
Bindings could not be implemented in `interfaces.hpp` such as suggested in #1091 because pybind11 generic lambda does not work with templates.
```
error: 'decltype' cannot resolve address of overloaded function
```
I had to instead implement within gqcpy/src/QCModel/HF/RHF_bindings.cpp directly. This will imply however that the binding implementation is currently limited to RHF and double (non-imaginary) basis sets.

@ThiboVE are you able to bugfix using this implementation, or does your issue only arise in complex basis sets? If so, I will implement a binding that works in complex scalar basises also.

emielvb:
@ThiboVE bindings have been successfully implemented in `RHF_d` and `RHF_cd`. Please review and merge if all github actions pass successfully.

ThiboVE (approved):

